PR #6, opened 2018-08-03
Enable std by default in pbkdf2

equal-l2:


equal-l2:
After second thought, I found current implementation of PBKDF2 didn't need std feature at all.
(I needed this PR for my subsequent PR)

PR #7, opened 2018-08-03
Port utility functions for pbkdf2 from rust-crypto

equal-l2:
Ported `pbkdf2_simple` and `pbkdf2_check`, and corresponding tests and docs from [`rust-crypto`](https://github.com/DaGenix/rust-crypto/blob/master/src/pbkdf2.rs).

### DIFF
--- a/pbkdf2/src/lib.rs
+++ b/pbkdf2/src/lib.rs
@@ -1,4 +1,4 @@
-#![cfg_attr(not(feature = "std"), no_std)]
+#![cfg_attr(not(feature = "include_simple"), no_std)]
 #![cfg_attr(feature = "cargo-clippy", allow(inline_always))]
 extern crate crypto_mac;
 extern crate generic_array;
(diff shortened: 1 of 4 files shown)


pbkdf2/src/lib.rs
```diff
@@ -74,3 +99,140 @@ pub fn pbkdf2<F>(password: &[u8], salt: &[u8], c: usize, res: &mut [u8])
...
+ * returns true if the passed in password hashes to the same value.
+ *
+ * # Arguments
+ *
+ * * password - The password to process as a str
+ * * hashed_value - A string representing a hashed password returned by pbkdf2_simple()
+ *
+ */
+#[cfg(feature="include_simple")]
+pub fn pbkdf2_check(password: &str, hashed_value: &str) -> Result<bool, &'static str> {
```
newpavlov: Can you please use the same API as in the `scrypt` crate, i.e. use `Result<(), CheckError>`? In future we will probably move these function to a trait, but for now it would be nice to keep API consistent across crates.

equal-l2: Done

newpavlov:
This PR is published (with some minor changes) as `pbkdf2 v0.2.1`.

PR #8, opened 2018-08-15
Bug in pbkdf2 xor function

brycx:
Currently the XOR function checks if the salt lenght is greater than or equal to itself:
```rust
[inline(always)]
fn xor(res: &mut [u8], salt: &[u8]) {
    assert!(salt.len() >= salt.len(), "length mismatch in xor");
    for i in 0..res.len() {
        res[i] ^= salt[i];
    }
}
```
I believe it actually should be checking that it's greater than or equal to `res`. Currently this doesn't show up in tests because the salt value is always passed with an `F::OutputSize` length and `chunk` will always be <= `F::OutputSize`, but I think this should be fixed in case the implementation changes in the future. I've checked the versions available on crates.io(0.1.0, 0.2.0, 0.2.1), too see if they all are implemented in a way that `chunk` will never be greater in length than a salt value passed. I found no problems with these three versions, but they all have this exact `assert!` line. I have also added regression tests to this PR.

### DIFF
--- a/pbkdf2/src/lib.rs
+++ b/pbkdf2/src/lib.rs
@@ -54,10 +54,9 @@ pub use errors::CheckError;
 
 #[inline(always)]
 fn xor(res: &mut [u8], salt: &[u8]) {
-    assert!(salt.len() >= res.len(), "length mismatch in xor");
-    for i in 0..res.len() {
-        res[i] ^= salt[i];
-    }
+    debug_assert!(salt.len() >= res.len(), "length mismatch in xor");
+
+    res.iter_mut().zip(salt.iter()).for_each(|(a, b)| *a ^= b);
 }
 
 #[inline(always)]
--- a/pbkdf2/src/lib.rs
+++ b/pbkdf2/src/lib.rs
@@ -54,7 +54,7 @@ pub use errors::CheckError;
 
 #[inline(always)]
 fn xor(res: &mut [u8], salt: &[u8]) {
-    assert!(salt.len() >= salt.len(), "length mismatch in xor");
+    assert!(salt.len() >= res.len(), "length mismatch in xor");
     for i in 0..res.len() {
         res[i] ^= salt[i];
     }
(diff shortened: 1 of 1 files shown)


newpavlov:
Oh, thank you for catching this! I think we can use `debug_assert!` here and change content of `xor` function to:
```Rust
res.iter_mut().zip(salt.iter()).for_each(|(a, b)| *a ^= b)
```

brycx:
Glad if I can help out! `debug_assert!` here is smart, suggested changes applied.

newpavlov:
BTW I've missed it, but you needed to use `byteorder` with disabled default features, for it to be `no_std` compatible. I'll fix it and will add tests.

brycx:
> you needed to use byteorder with disabled default features, for it to be `no_std` compatible

I use `byteorder` in another library, explicitly marked `no_std`, with default not disabled. Must have been introduced in version 1.2.5 or 1.2.6. Apologies for missing that, hope it's not too inconvenient.

newpavlov:
Ups, it was a wrong PR. :)

>Must have been introduced in version 1.2.5 or 1.2.6. Apologies for missing that, hope it's not too inconvenient.

Yeah, no problem. But note that `byteorder` had default `std` feature since [v0.5.2](https://github.com/BurntSushi/byteorder/blob/e3ce85194a6fffc758428c9ad26b8f01e4054c21/Cargo.toml).

brycx:
Lol yeah, haha.
> But note that byteorder had default std feature since v0.5.2.

I checked and saw that. I'll update my lib to be on the safe side. Thanks for pointing this out!

brycx:
Wait, should `rand` in `pbkdf2` and `scrypt` not also be imported with default set to false? Currently it's:
`rand = { version = "0.5", optional = true }`
But the docs say:
>no_std mode is activated by setting default-features = false

newpavlov:
We use `rand` only with enabled `std` feature, plus I think `OsRng` will not be available if we disable it.

brycx:
> plus I think OsRng will not be available if we disable it.

Yup. Thanks again for clarifying.

PR #10, opened 2018-08-30
Update byte-tools to 0.3 and add byteorder crate

brycx:
Updates to `byte-tools = 0.3` and adds `byteorder = 1.2` for endianness conversion.


newpavlov:
Thank you!

PR #12, opened 2018-09-14
Don't use rand::OsRng when compiling for wasm

AndyBitz:
When compiling for `wasm32-unknown-unknown` it will abort because of `rand::OsRng`.

```bash
error[E0432]: unresolved import `rand::OsRng`
  --> password-hashing/pbkdf2/src/lib.rs:42:12
   |
42 | use rand::{OsRng, RngCore};
   |            ^^^^^ no `OsRng` in the root
```

Since I only need `pbkdf2` and not `pbkdf2_simple`, which relies on `rand::OsRng`, I've just "configed" it and the `use` out for wasm.

However, I'm not sure if this is an actual fix or just a workaround for something else that needs to get fixed in order for this to work with wasm.

### DIFF
--- a/pbkdf2/src/lib.rs
+++ b/pbkdf2/src/lib.rs
@@ -38,7 +38,7 @@ use std::io;
 
 #[cfg(feature="include_simple")]
 use constant_time_eq::constant_time_eq;
-#[cfg(feature="include_simple")]
+#[cfg(all(not(target_arch="wasm32"), feature="include_simple"))]
 use rand::{OsRng, RngCore};
 #[cfg(feature="include_simple")]
 use hmac::Hmac;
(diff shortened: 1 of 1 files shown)


newpavlov:
As was written in the crate documentation you can simply disable default features for `pbkdf2`. To do it in your `Cargo.toml` use:
```toml
[dependencies]
pbkdf2 = { version = "0.2", default-features = false }
```

AndyBitz:
Ah, yeah, right at the top of the file.
I didn't know about features until now.

Thanks for your help.

newpavlov:
You also can use [docs.rs](https://docs.rs/pbkdf2) which is the de-facto standard documentation site for crates.

PR #13, opened 2018-12-14
Use subtle v2

hdevalence:
Hi, we released version 2 of `subtle` (context: https://github.com/RustCrypto/traits/pull/33#issue-238854180 ).

As far as I can tell, `subtle` is only used for `ConstantTimeEq`, which is only used internally, so this dependency upgrade will not be a semver breaking change.


newpavlov:
Yes, in this case we can update `subtle` without any problems. Though before merging I probably should understand why `thumbv7em-none-eabi` CI test fails. (at first glance looks like a nightly breakage)

Issue #17, opened 2019-07-01
Upgrade to rand 0.7.0


fubuloubu:
Using this package alongside a project that uses the latest `rand` crate, and getting an incompatibility:
```rust
   Compiling pbkdf2 v0.3.0
error[E0432]: unresolved import `rand::OsRng`
 --> ~/.cargo/registry/src/github.com-1ecc6299db9ec823/pbkdf2-0.3.0/src/simple.rs:7:12
  |
7 | use rand::{OsRng, RngCore};
  |            ^^^^^ no `OsRng` in the root
```

Solution: just upgrade the imports

breard-r:
The `rand` dependency is required only for the `include_simple` feature. If you do not use the simple function then, in your project, you can specify `default-features = false` when including `pbkdf2`.

PR #20, opened 2019-08-18
Refactoring of the `pbkdf2` crate.

jfrimmel:
This PR applies refactoring to the `pbkdf2` crate, especially for the `simple` module.
It is best viewed commit by commit.

### DIFF
--- a/pbkdf2/src/simple.rs
+++ b/pbkdf2/src/simple.rs
@@ -80,13 +80,13 @@ pub fn pbkdf2_check(password: &str, hashed_value: &str) -> Result<(), CheckError
         _ => Err(CheckError::InvalidFormat)?,
     };
 
-    let count = base64::decode(count).map_err(|_| CheckError::InvalidFormat)?;
+    let count = base64::decode(count)?;
     if count.len() != 4 {
         Err(CheckError::InvalidFormat)?;
     }
     let count = BigEndian::read_u32(&count[..]) as usize;
-    let salt = base64::decode(salt).map_err(|_| CheckError::InvalidFormat)?;
-    let hash = base64::decode(hash).map_err(|_| CheckError::InvalidFormat)?;
+    let salt = base64::decode(salt)?;
+    let hash = base64::decode(hash)?;
 
     let mut output = vec![0u8; hash.len()];
     pbkdf2::<Hmac<Sha256>>(password.as_bytes(), &salt, count, &mut output);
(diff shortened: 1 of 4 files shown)


jfrimmel:
Do you want to keep the minimum Rust version of 1.22? Some dependencies (e.g. rand) could also be updated, if the MSRV is bumped

pbkdf2/src/simple.rs
```diff
@@ -69,65 +70,26 @@ pub fn pbkdf2_simple(password: &str, c: u32) -> io::Result<String> {
...
-            Ok(salt) => salt,
-            Err(_) => return Err(CheckError::InvalidFormat)
-        },
-        None => return Err(CheckError::InvalidFormat)
+pub fn pbkdf2_check(password: &str, hashed_value: &str) -> Result<(), CheckError> {
+    let parts: Vec<_> = hashed_value.split('$').collect();
+
+    // check the format of the input: there may be no tokens before the first
+    // and after the last `$`, tokens must have correct information and length.
+    let (count, salt, hash) = match parts[..] {
```
newpavlov: I think you can remove the allocation by using:
```rust
let mut p = hashed_value.split('$');
let buf: [Option<&str>; 8] = [
    p.next(), p.next(), p.next(), p.next(),
    p.next(), p.next(), p.next(), p.next(),
];
let (count, salt, hash) = match buf {
    [
        Some(""), Some("rpbkdf2"), Some("0"), Some(count),
        Some(salt), Some(hash), Some(""), None,
    ] => (count, salt, hash),
    _ => return Err(CheckError::InvalidFormat),
};
```

Also it looks like `Err(CheckError::InvalidFormat)?` is discouraged (e.g. by clippy), so we probably better to replace it with an explicit `return`.

jfrimmel: Thank you for your suggestion. I will remove the unnecessary allocation (I wasn't happy with it, too) by applying your changes :smile: 
Unfortunately, I don't think, that Rust 1.22 will support this code as well.

jfrimmel:
> This will fix #17, correct?

Unfortunately no, since this PR does not increase the MSRV. This would be required to use Rand 0.7.

tarcieri:
@jfrimmel can you rebase? we've bumped MSRV to 1.41+

newpavlov:
I've included your refactor into #33.

PR #23, opened 2020-01-03
bcrypt-pbkdf crate

str4d:
Implements `bcrypt_pbkdf` as used in OpenSSH.


str4d:
Hmm... there are three CI errors:

- A `zeroize` compilation error which I think is caused by its MSRV being 1.39 while the MSRV of the whole repository is 1.22.
- An `scrypt` compilation error that I have no idea why fails here but not in master.
- A `zeroize` compilation error caused by `zeroize` being compiled in a `no_std` context (which I thought I'd disabled by excluding `bcrypt-pbkdf` from the `build_nostd.sh` script).

newpavlov:
Thank you! Although I am not sure it's correct to use the `Mac` trait here. I think instead we could just expose two separate functions, as done in the NPM crate. We may add a dedicated trait for password hashing functions sometime in the future, but now plain functions work just fine.

I will try to look into this PR and CI issues more thoroughly a bit later. Meanwhile could you please do the following:
- Move tests from `lib.rs` into a dedicated `tests/` folder.
- Share the crate ownership with me or @tarcieri.

str4d:
> Although I am not sure it's correct to use the Mac trait here. I think instead we could just expose two separate functions, as done in the NPM crate.

I'm not using the `Mac` trait as part of the public API. This crate simply exposes a single function for `bcrypt_pbkdf`; the `Bhash` struct is internal.

`Mac` is pulled in because the `pbkdf2` crate requires it for its own API, and I specifically wanted to build this crate around the `pbkdf2` crate.

> Meanwhile could you please do the following:

Will do!

newpavlov:
>Mac is pulled in because the pbkdf2 crate requires it for its own API

Ah, I see. For some reason I thought you expose `Bhash` in public API.

str4d:
> - Move tests from lib.rs into a dedicated tests/ folder.

I've moved the OpenBSD test vectors to the `tests/` folder. I can't move the `bhash` tests into that folder because the `bhash` function is completely internal.

> - Share the crate ownership with me or @tarcieri.

It looks like we've now moved to having GitHub teams added for ownership. What team should I add?

tarcieri:
I made a @RustCrypto/password-hashes team

str4d:
I've added the team as an owner of https://crates.io/crates/bcrypt-pbkdf

tarcieri:
It looks like this crate requires `std` (or at best, `alloc`) at present and the `no_std` build is failing?

Perhaps we should update this repo to GitHub Actions too.

str4d:
I have more changes locally that migrate this crate to `crypto-mac 0.8` etc ahead of an 0.2 release. I'll make a separate PR with them once this PR is merged, because they will depend on the bumped MSRV for this repository, which this PR doesn't have.

str4d:
@tarcieri said to just push the 0.2 changes to this PR. I will do that, and publish and tag 0.2, in this PR.

str4d:
v0.2.0 has been tagged and published, and this PR is passing tests and should be ready to merge.

bcrypt-pbkdf/tests/test_vectors.rs
```diff
@@ -0,0 +1,76 @@
+extern crate bcrypt_pbkdf;
```
tarcieri, line 1: This is probably no-longer needed

PR #28, opened 2020-06-04
Easy creation of recommended ScryptParam

liamrosenfeld:
This makes it easier to use by moving it to its own function.


newpavlov:
Thank you!

PR #40, opened 2020-07-03
Remove superflous main() in documentation

slightknack:
The current documentation has an unneeded `fn main() {}` in the *Usage* section. This commit removes the line; alternatively, the code example could be wrapped in `main`.

### DIFF
--- a/scrypt/src/lib.rs
+++ b/scrypt/src/lib.rs
@@ -27,6 +27,7 @@
 //! assert!(scrypt_check("Not so secure password", &hashed_password).is_ok());
 //! # }
 //! # #[cfg(not(feature="include_simple"))]
+//! # fn main() {}
 //! ```
 //!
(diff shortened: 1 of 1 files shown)


slightknack:
It seems to me now that the line I removed was a part of a doctest - is there a way to include the doctest without showing `fn main() {}` in the documentation?

slightknack:
I cloned the branch and rebuilt the documentation successfully. I'm not sure why the checks are failing.

slightknack:
The build fails on target `powerpc64-unknown-linux-gnu` while compiling `pbkdf2`. This seems to be out of the scope of this PR; perhaps an issue should be opened?

newpavlov:
Thank you! Looks like `cross` no longer provides the docker image which we were using. Either way, we probably should migrate to GIthub Actions.

PR #42, opened 2020-07-03
Fix the Travis CI badge

breard-r:


newpavlov:
Thank you for noticing! We've migrated from Travi CI to GitHub Actions, but I forgot to remove that badge. It's fixed now.

PR #44, opened 2020-08-18
pbkdf2: bump `crypto-mac` dependency to v0.9

tarcieri:


newpavlov:
I think it would be nice to bump `pbkdf2` minor version, update the dependent crates and cut releases directly after merging this PR.

tarcieri:
@newpavlov yup, that's the plan. We need to do `scrypt` too, as it depends on `pbkdf2`.

newpavlov:
I meant as part of this PR, but oh well. :)

tarcieri:
Aah, oh well! Preparing a follow-up PR now.

PR #74, opened 2020-12-28
sha-crypt: rename `simple` feature + doc_cfg

tarcieri:
...from `include-simple`. Just `simple` is... simpler.

Also adds `doc_cfg` for https://docs.rs to document functions gated on this feature.


tarcieri:
I now see `include_simple` is more consistent. Never mind.

Will re-add `doc_cfg` in a separate PR.

PR #82, opened 2021-01-11
pbkdf2: add PHC hash support using `password-hash` crate

tarcieri:
Implements the PHC string format for hashes, currently gated under the `include_simple` cargo feature.

Specifics of how the format is encoded/decoded are modeled off of the implementation of PHC hashes for PBKDF2 implemented by Auth0.

See this thread for more information:

https://community.auth0.com/t/bulk-user-import-custom-password-hash-issue/47408/5

### DIFF
--- a/.github/workflows/bcrypt-pbkdf.yml
+++ b/.github/workflows/bcrypt-pbkdf.yml
@@ -22,7 +22,7 @@ jobs:
     strategy:
       matrix:
         rust:
-          - 1.41.0 # MSRV
+          - 1.47.0 # MSRV
           - stable
         target:
           - thumbv7em-none-eabi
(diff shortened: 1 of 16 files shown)


tarcieri:
Note: based on actually using these traits, before I mark this as non-draft there are some refinements I'd like to do to the `password-hash` API which were closer to some of @newpavlov's original suggestions.

Namely I'd like to make a `ParamsStr` type which validates and wraps the original param string and provides iteration methods (sort of like `std::path::Path`), and an associated `type Params` to `PasswordHasher`, handling conversion to/from the algorithm's native `Params` format.

tarcieri:
Note: I updated this to incorporate https://github.com/RustCrypto/traits/pull/477 and it seems better now

PR #87, opened 2021-01-20
Argon2

tarcieri:
Translation of the CC0-licensed reference implementation as implemented in C, as of this commit:

commit 440ceb9612d5a20997e3e12728542df2de713ca4
Merge: 3df7b84 6d86209
Author: Dmitry Khovratovich <khovratovich@gmail.com>
Date:   Thu Jul 9 14:56:35 2020 +0200

This translation is based on the reference implementation (`ref.c`) as opposed to the optimized implementation (`opt.c`) to get to a baseline implementation which passed the test vectors, but it would probably be best to translate the optimized implementation.

Multithreading is also unimplemented.

Otherwise, (if it were working) this implementation provides Argon2d, Argon2i, and Argon2id along with support for versions `0x10` and `0x13`.


tarcieri:
I chose to do a translation of the C source code instead of adapting one of the existing implementations for this reasons:

- `no_std` compatibility: the C source code almost completely avoids heap allocations except for the instance "memory", which can be converted to a slice
- `blake2` crate: the other existing implementations don't use the [`blake2`](https://github.com/RustCrypto/hashes/tree/master/blake2) crate. This implementation is integrated with it.
- `password-hash` crate: Argon2 is a foremost use case for the PHC string format, and the [`password-hash`](https://github.com/RustCrypto/traits/tree/master/password-hash) crate provides a `no_std` friendly implementation of it, which is a key part of the first point.

tarcieri:
Looking at other Rust implementations of Argon2, [`argon2rs`](https://github.com/bryant/argon2rs) provides a much more idiomatic implementation which is also immaculately commented.

It'd be nice to repurpose `argon2rs` for our implementation, but the only problem is it's MIT-only as opposed to RustCrypto's standard dual-license MIT+Apache2.0 (following conventional Rust licensing policies).

@bryant how would you feel about allowing us to incorporate portions of `argon2rs`, but dual licensed as MIT+Apache 2.0

tarcieri:
This implementation now successfully computes the pre-hashing digest for an Argon2d v19 test vector.

The computed output is wrong, however.

argon2/src/lib.rs
```diff
@@ -0,0 +1,901 @@
...
+/// designed by the Lyra PHC team
+fn fBlaMka(x: u64, y: u64) -> u64 {
+    let m = 0xFFFFFFFF;
+    let xy = Wrapping((x & m) * (y & m)) * Wrapping(2);
+    (Wrapping(x) + Wrapping(y) + xy).0
+}
+
+fn rotr64(w: u64, c: u64) -> u64 {
+    (w >> c) | (w << (64 - c))
+}
```
tarcieri, line 842: @newpavlov any ideas on how we could leverage the `blake2` crate for this?

newpavlov: Not sure, I will have to refresh my memory about BLAKE2 and familiarize myself with the Argon2 algorithm first.

Also I think that `blake2` rework is a long overdue. We probably should do it based on the `blake2b_simd` and `blake2s_simd` crates and use this chance to expose parts necessary for Argon2.

argon2/src/lib.rs
```diff
@@ -0,0 +1,906 @@
...
+    }
+
+    /// Function creates first 2 blocks per lane
+    fn fill_first_blocks(&mut self, blockhash: &[u8]) -> Result<()> {
+        let mut hash = [0u8; ARGON2_BLOCK_SIZE];
+
+        for l in 0..self.lanes {
+            // Make the first and second block in each lane as G(H0||0||i) or
+            // G(H0||1||i)
+            for i in 0u32..2u32 {
```
tarcieri, line 477: Finally tracked down the bug in my original translation. This function needed to fill the first two blocks, but it was only filling one because I accidentally wrote the loop as `0u32..1u32` (which is non-inclusive).

With that corrected, it's passing an Argon2d test vector locally! 🎉

argon2/src/lib.rs
tarcieri: Cool, well I guess we'll need to use something like this (I'll clean it up a bit more) but it'd be great to be able to leverage a SIMD implementation here

tarcieri:
Cool, it now passes an Argon2d test vector. 🎉 Turns out there was a very small indexing bug which I made a line note about.

I'll also add test vectors for Argon2i, Argon2id, and version 16 of all of the variants, then do some refactoring.

tarcieri:
Marking this as ready for review.

There are still some TODOs to cleanup, but otherwise it should be good to go.

argon2/src/lib.rs
```diff
@@ -0,0 +1,360 @@
...
+        if MAX_AD_LENGTH < ad.len() {
+            return Err(Error::AdTooLong);
+        }
+
+        let memory_blocks = (self.segment_length * self.lanes * SYNC_POINTS) as usize;
+
+        // Hashing all inputs
+        #[allow(unused_mut)]
+        let mut initial_hash = self.initial_hash(argon2_type, pwd, salt, ad, out);
+        let mut memory = vec![Block::default(); memory_blocks];
```
tarcieri, line 308: This is the only part of the code that relies on `alloc`.

It'd be pretty easy to extract a method that operates over a `&mut [Block]` (or a newtype wrapping it), but it'd probably be nice to encapsulate it instead of exposing `Block`, and if we do that, waiting for const generics would be nice.

So I left it out of the initial implementation, but it should be pretty easy to add retroactively.

PR #93, opened 2021-01-28
argon2: PHC string format support

tarcieri:
Uses the `password-hash` crate to implement PHC string format support


tarcieri:
This is basically ready, aside from test vectors for Argon2 PHC strings, which I'm unfortunately having some trouble finding.

PR #97, opened 2021-01-29
pbkdf2: deprecate old `include_simple` API

tarcieri:
This API is based on an MCF hash format (`$rpbkdf2$...`) which is unique to this crate and therefore not interoperable.

As of #82, the PHC string format is supported, which provides an interoperable way to serialize PBKDF2 password hashes.

It also implements the `McfHasher` trait which can be used to both verify old `$rpbkdf2$` MCF hashes as well as upgrade them to PHC hashes.

This commit deprecates the old MCF API in the hope of getting people to upgrade to the PHC format.

### DIFF
--- a/pbkdf2/src/lib.rs
+++ b/pbkdf2/src/lib.rs
@@ -28,6 +28,7 @@ mod hasher;
 mod simple;
 
 #[cfg(feature = "include_simple")]
+#[allow(deprecated)]
 pub use crate::{
     errors::CheckError,
     hasher::{Algorithm, Params, Pbkdf2},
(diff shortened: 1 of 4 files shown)


newpavlov (commented):
I think we can simply remove those methods and cut a new minor release.

tarcieri:
Works for me.

What do you think about renaming the `include_simple` feature to just `simple`?

newpavlov:
Do we need this feature in the first place? I thought MCF and PHC support will be automatically available via `password-hash`.

tarcieri:
It requires a bunch of additional dependencies which may or may not make sense depending on context, for example the `sha2` crate, but also `password-hash` and `base64ct` (currently it uses `base64`, but that's easily replaced)

tarcieri:
Opened #98 which removes the old "simple" API entirely

PR #122, opened 2021-02-01
[NOT FOR MERGE] sha-crypt: add Base64 test

tarcieri:
Tries to test encoding a crypt(3)-flavored Base64 against a test vector produced using the `base64` crate.

Not working?(!)

### DIFF
--- a/sha-crypt/src/b64.rs
+++ b/sha-crypt/src/b64.rs
@@ -52,6 +52,20 @@ pub fn decode(source: &[u8]) -> Vec<u8> {
 }
 
 mod tests {
+    #[test]
+    fn encode_vector() {
+        let raw = [
+            0, 16, 131, 16, 81, 135, 32, 146, 139, 48, 211, 143, 65, 20, 147, 81, 85, 151, 97, 150,
+            155, 113, 215, 159, 130, 24, 163, 146, 89, 167, 162, 154, 171, 178, 219, 175, 195, 28,
+            179, 251, 240, 1, 8, 49, 5, 24, 114, 9, 40, 179, 13, 56, 244, 17, 73, 53, 21, 89, 118,
+            25, 105, 183, 29, 121,
+        ];
+        let encoded = b"./0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnyz./0123456789ABCDEFGHIJKLMNOPQRSE";
+
+        let e = super::encode(&raw);
+        assert_eq!(e, encoded);
+    }
+
     #[cfg(feature = "simple")]
     #[test]
     fn test_encode_decode() {


tarcieri:
@awidegreen I'm curious where you got these test vectors from:

https://github.com/RustCrypto/password-hashes/blob/ddaeb13/sha-crypt/tests/mod.rs#L58-L100

I'm not able to reproduce the encoding that `b64.rs` is using with another implementation of crypt-flavored Base64

awidegreen:
I'm pretty sure that I verified the result against `dovecot pw` as that was the use case I had at the time. The implementation (incl the b64) is inspired by https://www.akkadia.org/drepper/SHA-crypt.txt which is more or less what is used in [glibc](https://sourceware.org/git/?p=glibc.git;a=blob;f=crypt/sha512-crypt.c;h=1ebe010a8b498911514ab85d62abeac6f088d019;hb=HEAD)

If I remember correctly (long time ago), I implemented the b64 myself as the std base64 crates didn't work for this use case.

tarcieri:
@awidegreen that's strange, as the alphabet here appears to match what I'd expect for `crypt(3)`-flavored Base64:

https://github.com/RustCrypto/password-hashes/blob/ddaeb13/sha-crypt/src/defs.rs#L4

Here's the `base64` crate's corresponding alphabet for reference:

https://github.com/marshallpierce/rust-base64/blob/4296732/src/lib.rs#L111

awidegreen:
@tarcieri I'll have look at the base64 crate again and see if that could be used instead. 
As for the failing test in this PR, the current implementation in `b64.rs` is obviously pretty dumb always expect as 64 bytes input for the encode, as this is what the sha512 hashing outputs.

tarcieri:
> I'll have look at the base64 crate again and see if that could be used instead.

We're using the `base64ct` crate for the other crates in this repo:

https://docs.rs/base64ct/0.2.0/base64ct/struct.Base64Crypt.html

Ideally we'd like to make it work with the `password-hash` crate (and possibly add [PHC String Format](https://github.com/P-H-C/phc-string-format/blob/master/phc-sf-spec.md) support):

https://docs.rs/password-hash/

...however this would be the first crate where we'd be trying to support the `crypt(3)` Base64 encoding with `password_hash::PasswordHash`.

tarcieri:
The test failure is no longer panicking and just diverging (left actual, right expected):

https://github.com/RustCrypto/password-hashes/pull/122/checks?check_run_id=1810555520#step:4:39

```
thread 'b64::tests::encode_vector' panicked at 'assertion failed: `(left == right)`
  left: `[54, 50, 53, 46, 69, 50, 110, 112, 84, 67, 77, 47, 77, 54, 54, 50, 70, 55, 53, 52, 88, 83, 77, 48, 99, 54, 55, 54, 71, 67, 80, 75, 98, 105, 77, 49, 115, 54, 56, 65, 72, 72, 106, 97, 102, 121, 77, 50, 55, 55, 80, 69, 73, 73, 110, 113, 106, 67, 78, 51, 78, 66, 81, 73, 74, 78, 53, 53, 110, 83, 78, 52, 100, 104, 84, 77, 75, 83, 57, 119, 47, 103, 78, 53, 116, 47]`,
 right: `[46, 47, 48, 49, 50, 51, 52, 53, 54, 55, 56, 57, 65, 66, 67, 68, 69, 70, 71, 72, 73, 74, 75, 76, 77, 78, 79, 80, 81, 82, 83, 84, 85, 86, 87, 88, 89, 90, 97, 98, 99, 100, 101, 102, 103, 104, 105, 106, 107, 108, 109, 110, 121, 122, 46, 47, 48, 49, 50, 51, 52, 53, 54, 55, 56, 57, 65, 66, 67, 68, 69, 70, 71, 72, 73, 74, 75, 76, 77, 78, 79, 80, 81, 82, 83, 69]`', sha-crypt/src/b64.rs:66:9
```

awidegreen:
@tarcieri 

Well, even though the alphabet of this `b64.rs` and the crypt(3) impl in `base64ct` is the same, the sequence how the bytes are encoded is certainly different. The `base64ct` implementation does it [sequentially](https://github.com/RustCrypto/utils/blob/585ebd99e18ff0ebe11f3dd9d467466b3b55d653/base64ct/src/encoding.rs#L183-L188) and the one here is based on Dreppers [definition](https://www.akkadia.org/drepper/SHA-crypt.txt):

```
  e) the base-64 encoded final C digest.  The encoding used is as
       follows:

                 111111111122222222223333333333444444444455555555556666
       0123456789012345678901234567890123456789012345678901234567890123
       ----------------------------------------------------------------
       ./0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz

       Each group of three bytes from the digest produces four
       characters as output:

         1. character: the six low bits of the first byte
         2. character: the two high bits of the first byte and the
                       four low bytes from the second byte
	 3. character: the four high bits from the second byte and
                       the two low bits from the third byte
         4. character: the six high bits from the third byte

       The groups of three bytes are as follows (in this sequence).
       These are the indices into the byte array containing the
       digest, starting with index 0.  For the last group there are
       not enough bytes left in the digest and the value zero is used
       in its place.  This group also produces only three or two
       characters as output for SHA-256 and SHA-512 respectively.

       For SHA-256:

         #3   #2   #1   <-- byte number in group

          0 - 10 - 20
         21 -  1 - 11
         12 - 22 -  2
          3 - 13 - 23
         24 -  4 - 14
         15 - 25 -  5
          6 - 16 - 26
         27 -  7 - 17
         18 - 28 -  8
          9 - 19 - 29
          * - 31 - 30


       For SHA-512:

         #3   #2   #1   <-- byte number in group

          0 - 21 - 42
         22 - 43 -  1
         44 -  2 - 23
          3 - 24 - 45
         25 - 46 -  4
         47 -  5 - 26
          6 - 27 - 48
         28 - 49 -  7
         50 -  8 - 29
          9 - 30 - 51
         31 - 52 - 10
         53 - 11 - 32
         12 - 33 - 54
         34 - 55 - 13
         56 - 14 - 35
         15 - 36 - 57
         37 - 58 - 16
         59 - 17 - 38
         18 - 39 - 60
         40 - 61 - 19
         62 - 20 - 41
          * -  * - 63
```

It seems that this is tailored for the sha-crypt - as all reference I found are pointing back to dreppers definition.

tarcieri:
Wow, ok then. Wasn't aware it was so significantly different. TIL.

Issue #125, opened 2021-02-07
argon2: can't find `SaltString::generate` from the example in the doc


Geobert:
Hi,

I'm trying to use `argon2` and following the example I'm encountering an issue with `SaltString::generate` which can't be found.

What am I missing here?

Thanks!

tarcieri:
Apologies, this was something of a gap in v0.1.0 API (which you can probably guess from the version number is an initial release).

I've released v0.1.1 (#127) which adds an enabled-by-default `rand` feature which should make it available.

Can you confirm it addresses your issue?

Geobert:
Yes it did! Thank you!

PR #126, opened 2021-02-07
argon2: add `rand` feature

tarcieri:
Enables the `rand_core` feature of `password-hash`.

Enabled-by-default.


tarcieri:
Note: this is intended to simplify #125.

PR #135, opened 2021-02-12
argon2: avoid salt-length related panic (closes #134)

### LINKED ISSUE #134: argon2: Panic with invalid salt
The following code:
```rs
fn main() {
    use argon2::PasswordHasher as _;
    let password_hasher = argon2::Argon2::default();

    let _ = password_hasher.hash_password(
        b"password1",
        None,
        None,
        argon2::Params::default(),
        argon2::password_hash::Salt::new("somesalt").unwrap(),
    );
}
```
panics with
```
thread 'main' panicked at 'internal error: entered unreachable code', /home/cr/.cargo/registry/src/github.com-1ecc6299db9ec823/argon2-0.1.2/src/lib.rs:583:30
```
Which is from this expression: https://github.com/RustCrypto/password-hashes/blob/25f5be0659db91df094a46625bc4800585bc9552/argon2/src/lib.rs#L574-L586, the actual error type here is `argon2::Error::SaltTooShort`

tarcieri:
Adds some additional advance checks that Argon2 parameters are within range, and returns errors in the event they are not.

This prevents a panic caused by what was previously an `unimplemented!` call when converting error types.

The occurrence of the panic in the first place speaks to overall deficiencies in the `password-hash` crate's error handling strategy.

It's further complicated by the awkwardness of the `password_hash::Output::init_with` API.

The code in this commit has TODOs for addressing the above, but it'd be good to open issues upstream on the `password-hash` crate about these problems, and ideally update the TODOs in the code with references to the issues.

cc @izik1

### DIFF
--- a/argon2/src/lib.rs
+++ b/argon2/src/lib.rs
@@ -571,6 +571,22 @@ impl PasswordHasher for Argon2<'_> {
         )
         .map_err(|_| HasherError::Params(ParamsError::InvalidValue))?;
 
+        if MAX_PWD_LENGTH < password.len() {
+            return Err(HasherError::Password);
+        }
+
+        if !(MIN_SALT_LENGTH..=MAX_SALT_LENGTH).contains(&salt_bytes.len()) {
+            // TODO(tarcieri): better error types for this case
+            return Err(HasherError::Crypto);
+        }
+
+        // Validate associated data (optional param)
+        if MAX_AD_LENGTH < ad.len() {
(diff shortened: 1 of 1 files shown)


tarcieri:
@izik1 unfortunately this doesn't give a good error for incorrect salt lengths (it's presently `HasherError::Crypto`), but at least it doesn't panic anymore.

I'll open some issues about improving the error handling story in `password-hash` v0.2, and in the meantime, cut another release with this fix.

PR #141, opened 2021-02-28
Enable password-hash/std feature

li-kai:
Enables `password-hash/std` feature, allowing crates like argon2 to throw error up from password-hash.

This is helpful in situations where you want to bubble up the error:

```sh
error[E0277]: the trait bound `HashError: StdError` is not satisfied
   --> auth/src/models.rs:104:47
    |
104 |     let parsed_hash = PasswordHash::new(&hash)?;
    |                                               ^ the trait `StdError` is not implemented for `HashError`
    |
    = note: required because of the requirements on the impl of `From<HashError>` for `anyhow::Error`
    = note: required by `std::convert::From::from`
```

### DIFF
--- a/argon2/Cargo.toml
+++ b/argon2/Cargo.toml
@@ -27,6 +27,7 @@ rand_core = { version = "0.6", features = ["std"] }
 [features]
 default = ["password-hash", "rand"]
 rand = ["password-hash/rand_core"]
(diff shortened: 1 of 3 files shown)


tarcieri:
Thank you! This seems like an oversight.

PR #147, opened 2021-03-24
Bump password-hash to 0.1.2 due to breaking change

trevor-crypto:
I had a use case where I was using the v0.7.3 of `pbkdf2`, and when dependabot bumped to v0.7.4 there was an error with compiling `pbkdf2`:

```
 if Base64::decode(count, &mut count_arr)?.len() != 4 {
    |                                                         ^ the trait `From<base64ct::Error>` is not implemented for `HasherError`
...
```
I found that this is due to my version of `password-hash` being stuck on v0.1.1. Once I did a `cargo update -p password-hash` all was resolved.

### DIFF
--- a/pbkdf2/Cargo.toml
+++ b/pbkdf2/Cargo.toml
@@ -17,7 +17,7 @@ crypto-mac = "0.10"
 rayon = { version = "1", optional = true }
 base64ct = { version = "1", default-features = false, optional = true }
 hmac = { version = "0.10", default-features = false, optional = true }
-password-hash = { version = "0.1", default-features = false, optional = true, features = ["rand_core"]  }
+password-hash = { version = "0.1.2", default-features = false, optional = true, features = ["rand_core"]  }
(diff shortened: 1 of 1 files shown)


tarcieri:
Aah, okay. This looks like an accidental SemVer breaking change I wasn't expecting while bumping `base64ct` (see #144).

I tried addressing it in an alternative way: yanking the v0.1.1 release so v0.1.2 is the only version published. If you'd like, I can merge this too.

trevor-crypto:
@tarcieri I see...Not sure how that happened then. Maybe the correct thing for me to have done was `cargo clean` first

It isn't a blocker anymore for me, so it's up to you to close or merge :relieved: Thanks!

tarcieri:
It's really an oversight on my part for pushing out a SemVer-breaking release. I believed all the usages of `base64ct` were internal to the crate, forgetting the error type was part of the public API.

Anyway, I think yanking the previous releases is sufficient to address this issue, so I'm going to close this.

PR #149, opened 2021-03-26
argon2: Add parallel lane processing

### LINKED ISSUE #103: argon2: parallel implementation
The `argon2` crate supports multiple lanes (the core parallelism construct used by Argon2) but does not yet support computing lanes in parallel using multiple threads.

See: https://github.com/RustCrypto/password-hashes/blob/00c1e6f/argon2/src/instance.rs#L118-L137

Corresponding parallel code in the reference implementation: https://github.com/P-H-C/phc-winner-argon2/blob/440ceb9/src/core.c#L291-L372

smallglitch:
Adds parallel processing for lanes 
Closes #103 

The parallelism is gated behind the feature `parallel`  

When the feature is activated, the `#![forbid(unsafe_code)]` gets downgraded to `#![deny(unsafe_code)]` due to unsafe usage ([here](https://github.com/smallglitch/password-hashes/blob/f06f51d04bae051dc9ae809f016a3c85b82dac3a/argon2/src/instance.rs#L127-L135))  
The `#![no_std]` flag is being disabled as well

### DIFF
--- a/argon2/src/instance.rs
+++ b/argon2/src/instance.rs
@@ -8,6 +8,12 @@ use blake2::{
     Blake2b, Digest, VarBlake2b,
 };
 
+#[cfg(feature = "parallel")]
+use std::{
+    mem,
+    thread::{self, JoinHandle},
+};
+
 #[cfg(feature = "zeroize")]
 use zeroize::Zeroize;
 
@@ -115,10 +121,71 @@ impl<'a> Instance<'a> {
         Ok(instance)
     }
 
+    /// Create multiple mutable references for the current instance, one for every lane
+    #[cfg(feature = "parallel")]
+    #[allow(unsafe_code)]
+    unsafe fn mut_self_refs(&mut self) -> Vec<&'static mut Instance<'static>> {
+        let lanes = self.lanes;
+        // This transmute can be skipped when a scoped threadpool is used (or when `spawn_unchecked()` gets stabilised)
+        let this = mem::transmute::<_, &mut Instance<'static>>(self);
+        let this: *mut Instance<'static> = this;
+
+        // Dereference the raw pointer multiple times to create multiple mutable references
+        (0..lanes).map(|_| &mut *(this)).collect()
+    }
+
+    #[cfg(feature = "parallel")]
+    fn fill_memory_blocks_par(&mut self) {
+        for r in 0..self.passes {
+            for s in 0..SYNC_POINTS {
+                // Allocate a vector for the join handles
+                let mut join_handles: Vec<JoinHandle<()>> =
+                    Vec::with_capacity(self.threads as usize);
+
+                // Safety: - All threads that receive a references will be joined before the item gets dropped
+                //         - All the read and write operations *shouldn't* overlap
+                #[allow(unsafe_code)]
+                let self_refs = unsafe { self.mut_self_refs() };
+
+                for (l, self_ref) in (0..self.lanes).zip(self_refs) {
+                    // Join the first thread if the current lane index matches/exceeds the maximum number of threads
+                    if l >= self.threads {
+                        let handle = join_handles.remove(0);
+                        handle.join().unwrap();
+                    }
+
+                    // Spawn the thread
+                    let handle = thread::spawn(move || {
+                        self_ref.fill_segment(Position {
+                            pass: r,
+                            lane: l,
+                            slice: s,
+                            index: 0,
+                        });
+                    });
+
+                    join_handles.push(handle);
+                }
+
+                // Join all remaining threads
+                for handle in join_handles {
+                    handle.join().unwrap();
+                }
+            }
+
+            // GENKAT note: this is where `internal_kat` would be called
+        }
+    }
+
     /// Function that fills the entire memory t_cost times based on the first two
     /// blocks in each lane
     fn fill_memory_blocks(&mut self) {
-        // TODO(tarcieri): multithread support
+        #[cfg(feature = "parallel")]
+        if self.threads > 1 {
+            self.fill_memory_blocks_par();
+            return;
+        }
+
         // Single-threaded version for p=1 case
         for r in 0..self.passes {
             for s in 0..SYNC_POINTS {
--- a/argon2/src/instance.rs
+++ b/argon2/src/instance.rs
@@ -9,9 +9,9 @@ use blake2::{
 };
 
 #[cfg(feature = "parallel")]
-use std::{
-    mem,
-    thread::{self, JoinHandle},
+use {
+    rayon::iter::{ParallelBridge, ParallelIterator},
+    std::mem,
 };
 
 #[cfg(feature = "zeroize")]
@@ -138,24 +138,15 @@ impl<'a> Instance<'a> {
     fn fill_memory_blocks_par(&mut self) {
         for r in 0..self.passes {
             for s in 0..SYNC_POINTS {
-                // Allocate a vector for the join handles
-                let mut join_handles: Vec<JoinHandle<()>> =
-                    Vec::with_capacity(self.threads as usize);
-
                 // Safety: - All threads that receive a references will be joined before the item gets dropped
                 //         - All the read and write operations *shouldn't* overlap
                 #[allow(unsafe_code)]
                 let self_refs = unsafe { self.mut_self_refs() };
 
-                for (l, self_ref) in (0..self.lanes).zip(self_refs) {
-                    // Join the first thread if the current lane index matches/exceeds the maximum number of threads
-                    if l >= self.threads {
-                        let handle = join_handles.remove(0);
-                        handle.join().unwrap();
-                    }
-
-                    // Spawn the thread
-                    let handle = thread::spawn(move || {
+                (0..self.lanes)
+                    .zip(self_refs)
+                    .par_bridge()
+                    .for_each(|(l, self_ref)| {
                         self_ref.fill_segment(Position {
                             pass: r,
                             lane: l,
@@ -163,14 +154,6 @@ impl<'a> Instance<'a> {
                             index: 0,
                         });
                     });
-
-                    join_handles.push(handle);
-                }
-
-                // Join all remaining threads
-                for handle in join_handles {
-                    handle.join().unwrap();
-                }
             }
 
             // GENKAT note: this is where `internal_kat` would be called
--- a/argon2/src/instance.rs
+++ b/argon2/src/instance.rs
@@ -36,6 +36,39 @@ struct Position {
     index: u32,
 }
 
+/// Structure containing references to the memory blocks
+struct Memory<'a> {
+    /// Memory blocks
+    data: &'a mut [Block],
+
+    /// Size of the memory in blocks
+    size: usize,
+}
+
+impl<'a> Memory<'a> {
+    /// Instantiate a new memory struct
+    fn new(data: &'a mut [Block]) -> Self {
+        let size = data.len();
+
+        Self { data, size }
+    }
+
+    /// Get a copy of the block
+    fn get_block(&self, idx: usize) -> Block {
+        self.data[idx]
+    }
+
+    /// Get a mutable reference to the block
+    fn get_block_mut(&mut self, idx: usize) -> &mut Block {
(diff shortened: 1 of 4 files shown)


tarcieri:
Thanks for implementing this.

I think it might make sense to use `rayon` to manage the thread pool, similar to what we have in the `pbkdf2` crate:

https://github.com/RustCrypto/password-hashes/blob/48e3c53681b1cfeb41e94fc09fda59012fbc9272/pbkdf2/src/lib.rs#L110-L123

argon2/src/instance.rs
```diff
@@ -115,10 +121,54 @@ impl<'a> Instance<'a> {
...
+    #[allow(unsafe_code)]
+    unsafe fn mut_self_refs(&mut self) -> Vec<&'static mut Instance<'static>> {
+        let lanes = self.lanes;
+        // This transmute can be skipped when a scoped threadpool is used (or when `spawn_unchecked()` gets stabilised)
+        let this = mem::transmute::<_, &mut Instance<'static>>(self);
+        let this: *mut Instance<'static> = this;
+
+        // Dereference the raw pointer multiple times to create multiple mutable references
+        (0..lanes).map(|_| &mut *(this)).collect()
+    }
```
tarcieri, line 135: Don't want to get too bogged down in this, but I'm curious if there's a safe borrow splitting strategy that can be used here instead

tarcieri: It seems like `self.lanes.iter_mut()` could also work here?

smallglitch: Hmm, I'm not sure. Looking at the specification, it seems like cross-lane memory access is required

> Mapping J1, J2 to the reference block index. The value l = J2 mod p determines the index of the lane
from which the block will be taken

https://github.com/RustCrypto/password-hashes/blob/e483ce59cf356e9fd1d18b42c61c17b637ff05ed/argon2/src/instance.rs#L245-L260

Off the top of my head I can only think of using either unsafe or a mutex to share the memory between the lanes
The mutex method could work but I'm not sure how much it would impact performance

tarcieri: Here's an example of how to do safe mutable borrow splitting:

```rust
fn split_borrows<T>(mut slice: &mut [T]) -> Vec<&mut T> {
    let mut result = vec![];
    
    while let Some((elem, rest)) = slice.split_first_mut() {
        result.push(elem);
        slice = rest;
    }
    
    result
}
```

smallglitch: (I was about to reply with the same one-liner to your tweet :) )

I'm sorry but I don't really understand how I should split the memory up to make this work  
Because, as far as I understand, that still unfortunately leaves us with the problem that every lane potentially needs access to every part of the memory 

The memory (even when converted from `&mut [Block]` to an `Vec<&mut Block>`) can't be modified because a vector like this would still only return an `&&mut Block` which can't be used to call a function that needs an `&mut Block` 
(which makes sense, allowing mutations on the underlying data through basically an immutable reference without some kind of lock is a bad idea)

Even though there shouldn't be data races because of how argon2 works, the compiler has no way of knowing that. 
To the compiler it looks we are trying to access memory that another thread could potentially mutate

This could be solved safely by using a mutex but could lead to contention when running the blake2 rounds 
(the `fill_bytes` function could probably be changed to locking the mutex itself instead of requiring a mutable reference to be present beforehand. This should get rid of most of the contention)

argon2/src/lib.rs
```diff
@@ -62,13 +62,14 @@
 //! [key derivation function]: https://en.wikipedia.org/wiki/Key_derivation_function
 //! [Password Hashing Competition]: https://www.password-hashing.net/
 
-#![no_std]
+#![cfg_attr(not(feature = "parallel"), no_std)]
```
tarcieri, line 65: Would suggest leaving this as-is:

```suggestion
#[no_std]
```

...and adding the following below the existing similar directive for `extern crate alloc`:

```rust
#[cfg(feature = "std")]
extern crate std;
```

argon2/src/instance.rs
tarcieri: Okay, sorry, I now realize this is a much thornier problem than I thought it was at first.

I guess I don't have any good suggestions offhand. We can start with this.

argon2/src/instance.rs
```diff
@@ -8,6 +8,13 @@ use blake2::{
     Blake2b, Digest, VarBlake2b,
 };
 
+#[cfg(feature = "parallel")]
+use {
+    alloc::vec::Vec,
+    core::mem,
```
tarcieri, line 14: Small nit, but since you have `std` in scope you could do:

```suggestion
    std::{mem, vec::Vec},
```

smallglitch: `std` shouldn't be in scope here, I didn't add the `extern crate std` with the last commit 
The need for the direct usage of `std` was a leftover from before we started using rayon

tarcieri: Aah ok, never mind then!

argon2/src/lib.rs
```diff
@@ -68,7 +68,8 @@
     html_logo_url = "https://raw.githubusercontent.com/RustCrypto/meta/master/logo.svg",
     html_favicon_url = "https://raw.githubusercontent.com/RustCrypto/meta/master/logo.svg"
 )]
-#![forbid(unsafe_code)]
+#![cfg_attr(not(feature = "parallel"), forbid(unsafe_code))]
+#![cfg_attr(feature = "parallel", deny(unsafe_code))]
```
tarcieri, line 72: I think it's fine to just change this to:

```suggestion
#![deny(unsafe_code)]
```

tarcieri (approved):
Cool, looks good to me.

I'll give @newpavlov a few days to comment in case he can think of a better solution to the mutable aliasing problem.

argon2/src/instance.rs
```diff
@@ -115,10 +122,54 @@ impl<'a> Instance<'a> {
...
+    #[cfg(feature = "parallel")]
+    #[allow(unsafe_code)]
+    unsafe fn mut_self_refs(&mut self) -> Vec<&'static mut Instance<'static>> {
+        let lanes = self.lanes;
+        // This transmute can be skipped when a scoped threadpool is used (or when `spawn_unchecked()` gets stabilised)
+        let this = mem::transmute::<_, &mut Instance<'static>>(self);
+        let this: *mut Instance<'static> = this;
+
+        // Dereference the raw pointer multiple times to create multiple mutable references
+        (0..lanes).map(|_| &mut *(this)).collect()
```
nikomatsakis, line 135: Under the unsafe code guidelines we've been considering, this would definitely be considered UB (or, at least, it would be if you go on to use these `&mut` later). Each time you create an `&mut`, you are effectively "stealing" the access permission to the memory from the previous one, and only one `&mut` can have that at a time.

nikomatsakis: What would be valid is if you were doing something like `&mut (*this + i)` -- i.e., creating `&mut references to disjoint areas in memory.

nikomatsakis: It would *also* be ok to return a `Vec<*mut Instance>` and then ensure that the access you do are disjoint -- but that is pushing some unsafety onto the users of the raw pointers later (which is correct and valid).

nikomatsakis: Consider, with the code as is:

You could, in safe code, create threads that access the same fields from these `&mut Instance` pointers, and each of those threads would (rightly) think that it has unique access.

argon2/src/instance.rs
```diff
@@ -115,10 +122,54 @@ impl<'a> Instance<'a> {
...
+                // Safety: - All threads that receive a references will be joined before the item gets dropped
+                //         - All the read and write operations *shouldn't* overlap
+                #[allow(unsafe_code)]
+                let self_refs = unsafe { self.mut_self_refs() };
+
+                (0..self.lanes)
+                    .zip(self_refs)
+                    .par_bridge()
+                    .for_each(|(l, self_ref)| {
+                        self_ref.fill_segment(Position {
```
nikomatsakis, line 151: What does `fill_segment` do?

nikomatsakis (changes requested):
I got intrigued by the twitter thread, as @bascule hoped, but I don't quite understand what this code is trying to do. :( More pointers would be helpful! I do see some problems with it as written, though.

tarcieri:
I can try to write a short synopsis of how the Argon2 paper describes parallel operation of the algorithm (mostly summarizing sections 3.2 and 3.3):

https://www.password-hashing.net/argon2-specs.pdf

The algorithm operates over a matrix of "blocks" consisting of:

- 𝒑 rows ("lanes"), where 𝒑 is the desired number of worker threads
- 𝒒 columns

Each lane is further subdivided into 𝑺 = 4 "slices" (in Argon2 terminology, and referred to as `SYNC_POINTS` in the code), and the intersection of a slice and a lane forms a *segment* of length 𝒒 / 𝑺.

Segments of the same slice are computed in parallel and therefore cannot reference each other. So from a memory model perspective, what we'd really like is to mutably borrow the values of a slice, partition them into segments, and give each thread access to those segments.

However, we also need to allow all of the worker threads to simultaneously borrow all of the other blocks which do not belong to the segments being operated on to reference as inputs.

This is the tricky part: the `fill_segment` operation can reference blocks from the current lane, or other lanes, but will not reference blocks from the same *segment* being operated on.

Having just written all of that down (thanks for rubber ducking if nothing else), I think I have a better idea of how to model this problem safely in Rust: slices (in the Argon2 sense) should be the core level of granularity in which the working "memory" is organized.

The main loop of the algorithm is over the slices. Provided I'm actually understanding this correctly, we can borrow one slice mutably at the time and the others immutably. The mutably borrowed slice can then be subdivided into a segment for each lane, given to the worker threads along with immutable references to all of the other slices.

I think a big part of what's making this so tricky right now is the memory consists of a contiguous `Vec<Block>`. I think that might still be fine for the backing storage, but perhaps we could mediate access to segments through another type that splits the borrows by Argon2 "slice", allowing one "slice" to be acted on mutably and the others referenced as inputs.

tarcieri:
I think a next step which might be helpful in general is to extract a `struct Memory` which borrows from a backing `[Block]` slice, pass that to `Instance::new` instead of the raw `&'a mut [Block]`, and provide a method on `Memory` for accessing blocks by their `Position`.

From there we can look at borrow splitting the backing buffer so `Memory` only holds 3 of the 4 "slices" at a time, and can be used to look up the "reference" blocks being used to fill a segment, but it would not have access to the slice being operated over (which would be borrowed mutably and split up into segments among the worker threads).

tarcieri (changes requested):
Also based on @nikomatsakis's comments I'm going to unapprove this for now

argon2/src/instance.rs
tarcieri: This is hopefully answered in my comments below, but it's effectively the minimum unit of work that a single worker thread can operate on at a given time.

It pseudorandomly samples reference inputs from other "slices" (in the Argon2 nomeclature) that aren't currently being worked on by other threads, and uses them to fill a particular "segment" with pseudorandom values (computed using a hash function).

argon2/src/instance.rs
smallglitch: > It would also be ok to return a Vec<*mut Instance> and then ensure that the access you do are disjoint -- but that is pushing some unsafety onto the users of the raw pointers later (which is correct and valid)

So wrapping the raw pointer to the instance inside an `AtomicPtr` could work as a band aid fix?

smallglitch:
This should fix the UB as every thread now dereferences the pointer itself

tarcieri:
@smallglitch nice! I think that's a start.

Do you want to mark the PR as ready for review?

smallglitch:
Sure, I wasn't sure if you'd be ok with an unsafe implementation 
If an unsafe implementation is ok, I think this should be fine

tarcieri (approved):
Using `unsafe` is fine for now. I can circle back on a safe implementation.

Thanks for extracting a `Memory` type.

tarcieri:
I opened #154 to track making the implementation safe

Issue #199, opened 2021-08-18
Implement std::error::Error for argon2::Error


LukeMathWalker:
Is there a reason for not implementing `std::error::Error` for `argon2::Error`? Would a PR in this direction be appreciated?

tarcieri:
It's an oversight. A PR would be accepted.

Note that you'll need to conditionally link `std`. Here's an example:

https://github.com/RustCrypto/password-hashes/blob/master/scrypt/src/lib.rs#L51-L52

lkarthee:
Would this work or anything needs to be added to commit ? (I will create a PR once changes are finalised)

 https://github.com/lkarthee/password-hashes/commit/f0e1c36d27a62b9fc1f40c83e4d54564d1c1b111

tarcieri:
The `argon2` crate itself doesn't link `std`, so you need to add the boilerplate to do so which I linked above

LukeMathWalker:
You beated me to it @lkarthee 😁

lkarthee:
zero to PR in rust @LukeMathWalker 😁

PR #200, opened 2021-08-21
argon2: add std error impl

lkarthee:
Implementing std::error::Error for argon2 as per discussion in  https://github.com/RustCrypto/password-hashes/issues/199


argon2/src/error.rs
```diff
@@ -1,5 +1,8 @@
 //! Error type
 
+#[cfg(feature = "std")]
+extern crate std;
```
tarcieri, line 4: This needs to be in `lib.rs` (below `extern crate alloc`)

tarcieri: I'll fix it up

PR #232, opened 2021-09-02
Balloon implementation

daxpedda:
See #1.

There are a lot of small open questions, but the really important ones that need to be answered are:
- [x] I don't know what `ints_to_block` is supposed to be (https://github.com/RustCrypto/password-hashes/pull/232#discussion_r700919220)
- [x] `to_int` is unclear (https://github.com/RustCrypto/password-hashes/pull/232#discussion_r700925129)
- [x] The paper uses a `||` operator to describe modifying the salt for parallel execution, I don't know what this is (https://github.com/RustCrypto/password-hashes/pull/232#discussion_r700934385)
- [x] `cnt` can overflow, the paper and the reference implementation say nothing about that (https://github.com/RustCrypto/password-hashes/pull/232#discussion_r700937948)
- [x] I'm hashing `u32` values, the reference implementation uses `i32` (I think), the paper says nothing about it (https://github.com/RustCrypto/password-hashes/pull/232#discussion_r700945238)
- [x] The paper doesn't specify a default configuration (https://github.com/RustCrypto/password-hashes/pull/232/files#r700980909)

For reference:
- The paper: https://eprint.iacr.org/2016/027.pdf
- Reference (actually they call it prototype) implementation: https://github.com/henrycg/balloon/

### DIFF
--- a/.github/workflows/workspace.yml
+++ b/.github/workflows/workspace.yml
@@ -16,7 +16,7 @@ jobs:
       - uses: actions/checkout@v1
       - uses: actions-rs/toolchain@v1
         with:
-          toolchain: 1.51.0 # MSRV
+          toolchain: 1.56.1 # MSRV
           components: clippy
           override: true
           profile: minimal
--- a/balloon/Cargo.toml
+++ b/balloon/Cargo.toml
@@ -13,6 +13,7 @@ readme = "README.md"
 
 [dependencies]
 digest = { version = "0.9", default-features = false }
+crypto-bigint = { version = "0.2", git = "https://github.com/daxpedda/crypto-bigint", branch = "nonzero-div-rem", features = ["generic-array"] }
 
 # optional dependencies
 password-hash = { version = "0.3", default-features = false, optional = true }
--- a/balloon/Cargo.toml
+++ b/balloon/Cargo.toml
@@ -20,6 +20,7 @@ password-hash = { version = "0.3", default-features = false, optional = true }
 rayon = { version = "1", optional = true }
 
 [dev-dependencies]
+hex-literal = "0.3"
 sha2 = "0.9"
 
 [features]
--- a/balloon/Cargo.toml
+++ b/balloon/Cargo.toml
@@ -20,6 +20,7 @@ password-hash = { version = "0.3", default-features = false, optional = true }
 rayon = { version = "1", optional = true }
 
 [dev-dependencies]
+nachonavarro-balloon = { git = "https://github.com/daxpedda/nachonavarro-balloon" }
 sha2 = "0.9"
 
 [features]
--- a/balloon/Cargo.toml
+++ b/balloon/Cargo.toml
@@ -8,7 +8,8 @@ documentation = "https://docs.rs/balloon-hash"
 repository = "https://github.com/RustCrypto/password-hashes/tree/master/balloon"
 keywords = ["crypto", "password", "hashing"]
 categories = ["cryptography", "no-std"]
-edition = "2018"
+edition = "2021"
+rust-version = "1.56.1"
 readme = "README.md"
 
 [dependencies]
--- a/balloon/Cargo.toml
+++ b/balloon/Cargo.toml
@@ -20,7 +20,6 @@ password-hash = { version = "0.3", default-features = false, optional = true }
 rayon = { version = "1", optional = true }
 
 [dev-dependencies]
-pyo3 = "0.15"
 sha2 = "0.9"
 
 [features]
--- a/balloon/Cargo.toml
+++ b/balloon/Cargo.toml
@@ -19,7 +19,6 @@ password-hash = { version = "0.3", default-features = false, optional = true }
 rayon = { version = "1", optional = true }
 
 [dev-dependencies]
-hex = "0.4"
 sha2 = "0.9"
 
 [features]
--- a/balloon/Cargo.toml
+++ b/balloon/Cargo.toml
@@ -13,7 +13,7 @@ readme = "README.md"
 
 [dependencies]
 digest = { version = "0.9", default-features = false }
-crypto-bigint = { version = "0.2", git = "https://github.com/RustCrypto/crypto-bigint", default-features = false, features = ["generic-array"] }
+crypto-bigint = { version = "0.2.10", default-features = false, features = ["generic-array"] }
 
 # optional dependencies
 password-hash = { version = "0.3", default-features = false, optional = true }
--- a/balloon/Cargo.toml
+++ b/balloon/Cargo.toml
@@ -20,7 +20,7 @@ password-hash = { version = "0.3", default-features = false, optional = true }
 rayon = { version = "1", optional = true }
 
 [dev-dependencies]
-nachonavarro-balloon = { git = "https://github.com/khonsulabs/nachonavarro-balloon" }
+pyo3 = "0.15"
 sha2 = "0.9"
 
 [features]
--- a/balloon/Cargo.toml
+++ b/balloon/Cargo.toml
@@ -1,0 +1,34 @@
+[package]
+name = "balloon-hash"
+version = "0.1.0" # Also update html_root_url in lib.rs when bumping this
+description = "Pure Rust implementation of the Balloon password hashing function"
+authors = ["RustCrypto Developers"]
+license = "MIT OR Apache-2.0"
+documentation = "https://docs.rs/balloon-hash"
+repository = "https://github.com/RustCrypto/password-hashes/tree/master/balloon"
+keywords = ["crypto", "password", "hashing"]
+categories = ["cryptography", "no-std"]
+edition = "2018"
+readme = "README.md"
+
+[dependencies]
+digest = { version = "0.9", default-features = false }
+
+# optional dependencies
+password-hash = { version = "0.3", default-features = false, optional = true }
+rayon = { version = "1", optional = true }
+
+[dev-dependencies]
+hex = "0.4"
+sha2 = "0.9"
+
+[features]
+default = ["alloc", "password-hash", "rand"]
+alloc = []
+parallel = ["rayon", "std"]
+rand = ["password-hash/rand_core"]
+std = ["alloc", "password-hash/std"]
+
+[package.metadata.docs.rs]
+all-features = true
+rustdoc-args = ["--cfg", "docsrs"]
--- a/balloon/Cargo.toml
+++ b/balloon/Cargo.toml
@@ -20,7 +20,7 @@ password-hash = { version = "0.3", default-features = false, optional = true }
 rayon = { version = "1", optional = true }
 
 [dev-dependencies]
-nachonavarro-balloon = { git = "https://github.com/daxpedda/nachonavarro-balloon" }
+nachonavarro-balloon = { git = "https://github.com/khonsulabs/nachonavarro-balloon" }
 sha2 = "0.9"
 
 [features]
--- a/balloon/Cargo.toml
+++ b/balloon/Cargo.toml
@@ -13,7 +13,7 @@ readme = "README.md"
 
 [dependencies]
 digest = { version = "0.9", default-features = false }
-crypto-bigint = { version = "0.2", git = "https://github.com/RustCrypto/crypto-bigint", features = ["generic-array"] }
+crypto-bigint = { version = "0.2", git = "https://github.com/RustCrypto/crypto-bigint", default-features = false, features = ["generic-array"] }
 
 # optional dependencies
 password-hash = { version = "0.3", default-features = false, optional = true }
--- a/balloon/Cargo.toml
+++ b/balloon/Cargo.toml
@@ -13,7 +13,7 @@ readme = "README.md"
 
 [dependencies]
 digest = { version = "0.9", default-features = false }
-crypto-bigint = { version = "0.2", git = "https://github.com/daxpedda/crypto-bigint", branch = "nonzero-div-rem", features = ["generic-array"] }
+crypto-bigint = { version = "0.2", git = "https://github.com/RustCrypto/crypto-bigint", features = ["generic-array"] }
 
 # optional dependencies
 password-hash = { version = "0.3", default-features = false, optional = true }
--- a/balloon/Cargo.toml
+++ b/balloon/Cargo.toml
@@ -13,7 +13,7 @@ readme = "README.md"
 
 [dependencies]
 digest = { version = "0.9", default-features = false }
-crypto-bigint = { version = "0.2.10", default-features = false, features = ["generic-array"] }
+crypto-bigint = { version = "0.3", default-features = false, features = ["generic-array"] }
 
 # optional dependencies
 password-hash = { version = "0.3", default-features = false, optional = true }
--- a/balloon/Cargo.toml
+++ b/balloon/Cargo.toml
@@ -13,7 +13,7 @@ rust-version = "1.56.1"
 readme = "README.md"
 
 [dependencies]
-digest = { version = "0.9", default-features = false }
+digest = { version = "0.10", default-features = false }
 crypto-bigint = { version = "0.3", default-features = false, features = ["generic-array"] }
 
 # optional dependencies
@@ -22,7 +22,7 @@ rayon = { version = "1", optional = true }
 
 [dev-dependencies]
 hex-literal = "0.3"
-sha2 = "0.9"
+sha2 = "0.10"
 
 [features]
 default = ["alloc", "password-hash", "rand"]
--- a/balloon/src/error.rs
+++ b/balloon/src/error.rs
@@ -11,10 +11,14 @@ pub type Result<T> = core::result::Result<T, Error>;
 /// Error type.
 #[derive(Copy, Clone, Debug, Eq, PartialEq)]
 pub enum Error {
+    /// Algorithm identifier invalid.
+    AlgorithmInvalid,
     /// Memory cost is too small.
     MemoryTooLittle,
     /// Not enough threads.
     ThreadsTooFew,
+    /// Too many threads.
+    ThreadsTooMany,
     /// Time cost is too small.
     TimeTooSmall,
 }
@@ -22,8 +26,10 @@ pub enum Error {
 impl fmt::Display for Error {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         f.write_str(match self {
+            Error::AlgorithmInvalid => "algorithm identifier invalid",
             Error::MemoryTooLittle => "memory cost is too small",
             Error::ThreadsTooFew => "not enough threads",
+            Error::ThreadsTooMany => "too many threads",
             Error::TimeTooSmall => "time cost is too small",
         })
     }
@@ -34,8 +40,10 @@ impl fmt::Display for Error {
 impl From<Error> for password_hash::Error {
     fn from(err: Error) -> password_hash::Error {
         match err {
+            Error::AlgorithmInvalid => password_hash::Error::Algorithm,
             Error::MemoryTooLittle => InvalidValue::TooShort.param_error(),
             Error::ThreadsTooFew => InvalidValue::TooShort.param_error(),
+            Error::ThreadsTooMany => InvalidValue::TooLong.param_error(),
             Error::TimeTooSmall => InvalidValue::TooShort.param_error(),
         }
     }
--- a/balloon/src/error.rs
+++ b/balloon/src/error.rs
@@ -1,0 +1,53 @@
+//! Error type
+
+use core::fmt;
+
+#[cfg(feature = "password-hash")]
+use password_hash::errors::InvalidValue;
+
+/// Result with balloon's [`Error`] type.
+pub type Result<T> = core::result::Result<T, Error>;
+
+/// Error type.
+#[derive(Copy, Clone, Debug, Eq, PartialEq)]
+pub enum Error {
+    /// Memory cost is too small.
+    MemoryTooLittle,
+    /// Output is too short.
+    OutputTooShort,
+    /// Not enough threads.
+    ThreadsTooFew,
+    /// Too many threads.
+    ThreadsTooMany,
+    /// Time cost is too small.
+    TimeTooSmall,
+}
+
+impl fmt::Display for Error {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.write_str(match self {
+            Error::MemoryTooLittle => "memory cost is too small",
+            Error::OutputTooShort => "output is too short",
+            Error::ThreadsTooFew => "not enough threads",
+            Error::ThreadsTooMany => "too many threads",
+            Error::TimeTooSmall => "time cost is too small",
+        })
+    }
+}
+
+#[cfg(feature = "password-hash")]
+#[cfg_attr(docsrs, doc(cfg(feature = "password-hash")))]
+impl From<Error> for password_hash::Error {
+    fn from(err: Error) -> password_hash::Error {
+        match err {
+            Error::MemoryTooLittle => InvalidValue::TooShort.param_error(),
+            Error::OutputTooShort => password_hash::Error::OutputTooShort,
+            Error::ThreadsTooFew => InvalidValue::TooShort.param_error(),
+            Error::ThreadsTooMany => InvalidValue::TooLong.param_error(),
+            Error::TimeTooSmall => InvalidValue::TooShort.param_error(),
+        }
+    }
+}
+
+#[cfg(feature = "std")]
+impl std::error::Error for Error {}
--- a/balloon/src/error.rs
+++ b/balloon/src/error.rs
@@ -15,8 +15,6 @@ pub enum Error {
     MemoryTooLittle,
     /// Not enough threads.
     ThreadsTooFew,
-    /// Too many threads.
-    ThreadsTooMany,
     /// Time cost is too small.
     TimeTooSmall,
 }
@@ -26,7 +24,6 @@ impl fmt::Display for Error {
         f.write_str(match self {
             Error::MemoryTooLittle => "memory cost is too small",
             Error::ThreadsTooFew => "not enough threads",
-            Error::ThreadsTooMany => "too many threads",
             Error::TimeTooSmall => "time cost is too small",
         })
     }
@@ -39,7 +36,6 @@ impl From<Error> for password_hash::Error {
         match err {
             Error::MemoryTooLittle => InvalidValue::TooShort.param_error(),
             Error::ThreadsTooFew => InvalidValue::TooShort.param_error(),
-            Error::ThreadsTooMany => InvalidValue::TooLong.param_error(),
             Error::TimeTooSmall => InvalidValue::TooShort.param_error(),
         }
     }
--- a/balloon/src/error.rs
+++ b/balloon/src/error.rs
@@ -13,8 +13,6 @@ pub type Result<T> = core::result::Result<T, Error>;
 pub enum Error {
     /// Memory cost is too small.
     MemoryTooLittle,
-    /// Output is too short.
-    OutputTooShort,
     /// Not enough threads.
     ThreadsTooFew,
     /// Too many threads.
@@ -27,7 +25,6 @@ impl fmt::Display for Error {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         f.write_str(match self {
             Error::MemoryTooLittle => "memory cost is too small",
-            Error::OutputTooShort => "output is too short",
             Error::ThreadsTooFew => "not enough threads",
             Error::ThreadsTooMany => "too many threads",
             Error::TimeTooSmall => "time cost is too small",
@@ -41,7 +38,6 @@ impl From<Error> for password_hash::Error {
     fn from(err: Error) -> password_hash::Error {
         match err {
             Error::MemoryTooLittle => InvalidValue::TooShort.param_error(),
-            Error::OutputTooShort => password_hash::Error::OutputTooShort,
             Error::ThreadsTooFew => InvalidValue::TooShort.param_error(),
             Error::ThreadsTooMany => InvalidValue::TooLong.param_error(),
             Error::TimeTooSmall => InvalidValue::TooShort.param_error(),
--- a/balloon/src/lib.rs
+++ b/balloon/src/lib.rs
@@ -139,8 +139,9 @@ impl<'key, D: Digest> Balloon<'key, D> {
 
                         // add thread index to salt
                         let byte = salt[..U32_SIZE].try_into().unwrap();
-                        salt[..U32_SIZE]
-                            .copy_from_slice(&(u32::from_le_bytes(byte).wrapping_add(index)).to_le_bytes());
+                        salt[..U32_SIZE].copy_from_slice(
+                            &(u32::from_le_bytes(byte).wrapping_add(index)).to_le_bytes(),
+                        );
                         Self::new(*params, *secret).hash(pwd, &salt)
                     })
                     .try_reduce(GenericArray::default, |a, b| {
--- a/balloon/src/lib.rs
+++ b/balloon/src/lib.rs
@@ -62,10 +62,12 @@
 #[cfg(feature = "alloc")]
 extern crate alloc;
 
+mod algorithm;
 mod error;
 mod params;
 
 pub use crate::{
+    algorithm::Algorithm,
     error::{Error, Result},
     params::Params,
 };
@@ -79,7 +81,10 @@ use digest::Digest;
 #[cfg_attr(docsrs, doc(cfg(feature = "password-hash")))]
 pub use password_hash::{self, PasswordHash, PasswordHasher, PasswordVerifier};
 #[cfg(all(feature = "alloc", feature = "password-hash"))]
-use password_hash::{Decimal, Ident, ParamsString, Salt};
+use {
+    core::convert::TryFrom,
+    password_hash::{Decimal, Ident, ParamsString, Salt},
+};
 
 /// Balloon context.
 ///
@@ -94,6 +99,8 @@ where
 {
     /// Storing which hash function is used
     pub digest: PhantomData<D>,
+    /// Algorithm to use
+    pub algorithm: Algorithm,
     /// Algorithm parameters
     pub params: Params,
     /// Key array
@@ -105,9 +112,10 @@ where
     GenericArray<u8, D::OutputSize>: ArrayDecoding,
 {
     /// Create a new Balloon context.
-    pub fn new(params: Params, secret: Option<&'key [u8]>) -> Self {
+    pub fn new(algorithm: Algorithm, params: Params, secret: Option<&'key [u8]>) -> Self {
         Self {
             digest: PhantomData,
+            algorithm,
             params,
             secret,
         }
@@ -140,11 +148,20 @@ where
         salt: &[u8],
         memory_blocks: &mut [GenericArray<u8, D::OutputSize>],
     ) -> Result<GenericArray<u8, D::OutputSize>> {
-        if self.params.p_cost.get() == 1 {
-            self.hash_internal(pwd, salt, memory_blocks, None)
-        } else {
+        if matches!(self.algorithm, Algorithm::Balloon) && self.params.p_cost.get() > 1 {
+            return Err(Error::ThreadsTooMany);
+        }
+
+        match self.algorithm {
+            Algorithm::Balloon => {
+                if self.params.p_cost.get() == 1 {
+                    self.hash_internal(pwd, salt, memory_blocks, None)
+                } else {
+                    Err(Error::ThreadsTooMany)
+                }
+            }
             #[cfg(not(feature = "parallel"))]
-            {
+            Algorithm::BalloonM => {
                 let mut result = GenericArray::default();
 
                 for thread in 1..=u64::from(self.params.p_cost.get()) {
@@ -155,7 +172,7 @@ where
                 Ok(result)
             }
             #[cfg(feature = "parallel")]
-            {
+            Algorithm::BalloonM => {
                 use rayon::iter::{ParallelBridge, ParallelIterator};
 
                 if memory_blocks.len()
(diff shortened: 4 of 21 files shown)


balloon/src/lib.rs
```diff
@@ -0,0 +1,346 @@
...
+                // for i from 0 to delta-1:
+                for i in 0..DELTA {
+                    // block_t idx_block = ints_to_block(t, m, i)
+                    // int other = to_int(hash(cnt++, salt, idx_block)) mod s_cost
+                    digest.update(&cnt.to_le_bytes());
+                    cnt += 1;
+                    digest.update(salt);
+                    digest.update(&u32::try_from(t).unwrap().to_le_bytes());
+                    digest.update(&u32::try_from(m).unwrap().to_le_bytes());
+                    digest.update(&i.to_le_bytes());
```
daxpedda: I don't know what `ints_to_block` is supposed to mean, the reference implementation does something with AES (I think) that I didn't understand: [libballoon/hash_state.c#L171](https://github.com/henrycg/balloon/blob/0ae35a8879a97f4c7c44ee7fe77ea6522c607ef7/libballoon/hash_state.c#L171) -> [libballoon/bitstream.c#L152](https://github.com/henrycg/balloon/blob/0ae35a8879a97f4c7c44ee7fe77ea6522c607ef7/libballoon/bitstream.c#L152) -> [libballoon/bitstream.c#L134](https://github.com/henrycg/balloon/blob/0ae35a8879a97f4c7c44ee7fe77ea6522c607ef7/libballoon/bitstream.c#L134) -> [libballoon/bitstream.c#L118](https://github.com/henrycg/balloon/blob/0ae35a8879a97f4c7c44ee7fe77ea6522c607ef7/libballoon/bitstream.c#L118) -> [EVP_EncryptUpdate](https://www.openssl.org/docs/man1.1.1/man3/EVP_EncryptUpdate.html)

I took some inspiration from [a Python implementation](https://github.com/nachonavarro/balloon-hashing/blob/5e1eb7acf008428b7cca96fa5dbb9a481bed2eef/balloon.py#L78) instead.

balloon/src/lib.rs
```diff
@@ -0,0 +1,347 @@
...
+                    digest.update(&cnt.to_le_bytes());
+                    cnt = cnt.wrapping_add(1);
+                    digest.update(salt);
+                    digest.update(&u32::try_from(t).unwrap().to_le_bytes());
+                    digest.update(&u32::try_from(m).unwrap().to_le_bytes());
+                    digest.update(&i.to_le_bytes());
+                    let other =
+                        u32::from_le_bytes(digest.finalize_reset()[..4].try_into().unwrap())
+                            % self.params.s_cost.get();
+                    let other = usize::try_from(other).unwrap();
```
daxpedda, line 241: This is supposed to be `to_int(..) mod s_cost`, instead of actually getting the remainder of the whole number, I'm only extracting the first byte. How should I approach this? I don't want to pull in num-bigint for this, crypto-bigint doesn't support this and I'm unsure if this really needs to be done in constant-time.

The only other solution I can come up with is to write a custom bigint remainder algorithm in here.

balloon/src/lib.rs
```diff
@@ -0,0 +1,347 @@
...
+use password_hash::{Decimal, Ident, ParamsString, Salt};
+
+/// Balloon context.
+///
+/// This is the primary type of this crate's API, and contains the following:
+///
+/// - Default set of [`Params`] to be used
+/// - (Optional) Secret key a.k.a. "pepper" to be used
+#[derive(Clone, Default)]
+pub struct Balloon<'key, D: Digest> {
```
daxpedda, line 89: I decided to store the hash algorithm to use in `Balloon` because I can't pass it into `PasswordHasher` otherwise.

I could also move it to `Params`, but I was unsure if I should store the hash algorithm in the PHC string and I don't know how to store it in the PHC string anyway, where do I get an ID or name for every hash algorithm from?

Another idea would be to make `PasswordHasher` always use SHA-256, or some other algorithm.

balloon/src/lib.rs
```diff
@@ -0,0 +1,347 @@
...
+                            new_salt.to_vec()
+                        } else {
+                            salt.to_vec()
+                        };
+
+                        // add thread index to salt
+                        let byte = salt[..U32_SIZE].try_into().unwrap();
+                        salt[..U32_SIZE].copy_from_slice(
+                            &(u32::from_le_bytes(byte).wrapping_add(index)).to_le_bytes(),
+                        );
```
daxpedda, line 144: This is what the paper proposes:
> BalloonM(p,s) := Balloon(p, s‖“1”) ⊕···⊕Balloon(p, s‖“M”)`

I don't know what `||` is, I found this: https://en.wikipedia.org/wiki/Parallel_(operator).
The reference implementation does this: [libballoon/balloon.c#L62-L76](https://github.com/henrycg/balloon/blob/0ae35a8879a97f4c7c44ee7fe77ea6522c607ef7/libballoon/balloon.c#L62-L76), but I didn't want to blindly translate some code.

balloon/src/lib.rs
```diff
@@ -0,0 +1,347 @@
...
+        }
+
+        let s_cost = usize::try_from(self.params.s_cost.get()).unwrap();
+        let t_cost = usize::try_from(self.params.t_cost.get()).unwrap();
+
+        let mut digest = D::new();
+
+        // This is a direct translation of the `Balloon` from <https://eprint.iacr.org/2016/027.pdf> chapter 3.1.
+        // int delta = 3 // Number of dependencies per block
+        const DELTA: u32 = 3;
```
daxpedda, line 180: Some implementations also make this configurable, should we do this too? What name should I give it in the PHC string?

balloon/src/lib.rs
```diff
@@ -0,0 +1,347 @@
...
+        let mut cnt: u32 = 0;
+        // block_t buf[s_cost]): // The main buffer
+        let memory_blocks = memory_blocks
+            .get_mut(..s_cost)
+            .ok_or(Error::MemoryTooLittle)?;
+
+        // Step 1. Expand input into buffer.
+        // buf[0] = hash(cnt++, passwd, salt)
+        digest.update(cnt.to_le_bytes());
+        cnt = cnt.wrapping_add(1);
```
daxpedda, line 191: I couldn't find what the reference implementation does against overflow for the counter, is wrapping alright here? The only other solution I can come up with is use a `u64` here, I would have to check, but it should suffice to cover all possibilities.

balloon/src/lib.rs
```diff
@@ -0,0 +1,347 @@
...
+                // for i from 0 to delta-1:
+                for i in 0..DELTA {
+                    // block_t idx_block = ints_to_block(t, m, i)
+                    // int other = to_int(hash(cnt++, salt, idx_block)) mod s_cost
+                    digest.update(&cnt.to_le_bytes());
+                    cnt = cnt.wrapping_add(1);
+                    digest.update(salt);
+                    digest.update(&u32::try_from(t).unwrap().to_le_bytes());
+                    digest.update(&u32::try_from(m).unwrap().to_le_bytes());
+                    digest.update(&i.to_le_bytes());
```
daxpedda, line 237: I am using `u32` here, which is arbitrary, as far as I could see the research paper says nothing about it. The reference implementation uses `i32` I believe, other implementations out there use `u32` and `u64`.

balloon/src/lib.rs
```diff
@@ -0,0 +1,347 @@
...
+    {
+        let salt = Salt::try_from(salt.as_ref())?;
+        let mut salt_arr = [0u8; 64];
+        let salt_bytes = salt.b64_decode(&mut salt_arr)?;
+
+        let output = password_hash::Output::new(&self.hash(password, salt_bytes)?)?;
+
+        Ok(PasswordHash {
+            algorithm: Ident::new("balloon"),
+            version: Some(1),
```
daxpedda, line 281: We don't have a standard or spec, it's "just" a research paper, it never mentions PHC strings. I just took those from the reference implementation.

balloon/src/lib.rs
```diff
@@ -0,0 +1,347 @@
...
+
+        let salt = salt.into();
+
+        Self::new(params, self.secret).hash_password(password, salt.as_str())
+    }
+}
+
+#[cfg(all(test, feature = "alloc", feature = "password-hash"))]
+#[test]
+fn hash_simple_retains_configured_params() {
```
daxpedda, line 316: I'm happy to add any other tests, but there are no test vectors out there as far as I could find.

Potentially I could compile the reference implementation as a dev-dependency (I don't know how much work that would be), but for that to be useful, it would require the implementation details to be exactly the same, which it already isn't, see all the notes above.

balloon/src/error.rs
```diff
@@ -0,0 +1,49 @@
...
+    TimeTooSmall,
+}
+
+impl fmt::Display for Error {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.write_str(match self {
+            Error::MemoryTooLittle => "memory cost is too small",
+            Error::ThreadsTooFew => "not enough threads",
+            Error::ThreadsTooMany => "too many threads",
+            Error::TimeTooSmall => "time cost is too small",
```
daxpedda, line 30: I could add more descriptive messages here. `MemoryTooLittle` could mention how much memory is actually needed. `ThreadsTooMany` could mention that you need the `parallel` crate feature to use threads. I don't know how to translate any of this to `password_hash::Error` though.

balloon/src/params.rs
```diff
@@ -0,0 +1,108 @@
...
+
+impl Params {
+    /// Default memory cost.
+    pub const DEFAULT_S_COST: u32 = 1024;
+
+    /// Default number of iterations (i.e. "time").
+    pub const DEFAULT_T_COST: u32 = 1;
+
+    /// Default degree of parallelism.
+    pub const DEFAULT_P_COST: u32 = 1;
```
daxpedda, line 32: I just used the defaults from the reference implementation, this is far from optimal though. Any guidance would be appreciated.

balloon/src/lib.rs
rozbb: Also not an expert, but ints_to_block seems to just be for type safety than an actual cryptographic operation. Feeding directly into the hash seems fine

balloon/src/lib.rs
rozbb: `s_cost` is an int64 in the prototype impl, so I'd say you should just do that and take the first 8 bytes from the digest

balloon/src/lib.rs
rozbb: A thought: the key size is fixed to the size of a `block_t`, which is the hash digest size. So instead of making `Balloon` depend on a `'key` lifetime, you can just make the secret a `GenericArray<u8, D::OutputSize>` or something

balloon/src/lib.rs
rozbb: || is the concatenation operation

balloon/src/params.rs
rozbb: I think you can get rid of the `new()` function and just leave the struct fields as public. The function doesn't do any bounds checking that isn't already enforced by the type system

balloon/src/lib.rs
rozbb: re the hasher: parameterizing by `D` and using a `PhantomData` is a pretty standard pattern

balloon/src/lib.rs
rozbb: I say fix delta. I'm not certain implementors will know what they're doing by tweaking it

balloon/src/lib.rs
rozbb: I think `cnt` and `i` should both be `u64`. If you had a 5GHz computer that did an increment every clock cycle it'd take over 100 years to overflow. Whereas with a `u32` it'd take less than a second.

balloon/src/lib.rs
rozbb: Agreed, this is a tough call. It's definitely risky to publish something that has no test vectors to speak of. @tarcieri what do you think?

rozbb:
Maybe some baseline questions to guide the discussion of whether to put this in this repo:
1. Does there exist any attempt to standardize this algorithm
2. Who uses this algorithm?
3. What are the alternatives to this algorithm? What are their pros and cons?

balloon/src/lib.rs
tarcieri: If there aren't test vectors to speak of, then at the very least I would suggest creating some with the reference implementation and leave the tool used to do so online somewhere.

If you could make a Rust FFI crate to so, and check it in as part of the initial implementation, that'd be great. We can potentially remove it from the source tree later, but it'd be nice to capture the provenance of the test vectors in a way that's reproducible and exists in the source code history.

tarcieri:
So far we've been fairly inclusive of algorithms which have some standardization effort behind them. We don't exactly have a policy regarding who's allowed to describe standards.

This algorithm does seem a bit nonstandard in that regard.

balloon/src/lib.rs
daxpedda: Alright then.

balloon/src/lib.rs
daxpedda: > What precisely would be the purpose of a "pepper" in this scheme? Honest question because I've never seen it used.

I thought the reason it was used in Argon2 was because it is considered good practice. See [OWASP](https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html#peppering) for example.

> re the hasher: parameterizing by `D` and using a `PhantomData` is a pretty standard pattern

Yes, I'm familiar with it, I'm just uncertain because which hash algorithm is used is quiet important, considering `Params` isn't storing it, it's being left out of the PHC string too. In any case, as noted in Zulip, we might wanna remove the whole `PasswordHasher` anyway.

balloon/src/lib.rs
daxpedda: I see, that could simply allow me to just add it to the hash, no reason to allocate.

balloon/src/lib.rs
daxpedda: I don't mind, but that would require us to properly mirror the prototype implementation, which would lead us back to https://github.com/RustCrypto/password-hashes/pull/232#discussion_r700919220 and `EVP_EncryptUpdate` which is still have little clue what it does, but if it involves some sort of encryption, I think this isn't what we want, right?

tarcieri: `EVP_EncryptUpdate` is a (now deprecated) OpenSSL API:

https://www.openssl.org/docs/manmaster/man3/EVP_EncryptUpdate.html

You said it was doing something with AES... is it possible it was deriving an AES key?

The test vectors should probably cover the raw output. Worst case you could add a public API for that so you don't have to go through an OpenSSL compatibility layer.

Mirroring the sources is fine if they aren't too large. It'd just be nice to have a safe place where the test vectors can be generated/regenerated.

daxpedda:
> 1. Does there exist any attempt to standardize this algorithm

Not as far as I know, and the research paper really leaves a lot to desire that is usually expected in a standard, which is to be expected.
> 2. Who uses this algorithm?

I couldn't find anybody who does. The biggest motivation I found is that it is [the only alternative mentioned by NIST to PBKDF2](https://pages.nist.gov/800-63-3/sp800-63b.html#memsecretver). 
> 3. What are the alternatives to this algorithm? What are their pros and cons?

Almost half of the research paper compares it to Argon2. I'm no cryptographer, but from what I could make out the biggest advantages are:
- Simplicity
- Formal verification (of it's memory-hardness capabilities?)
- Speed

To me the biggest disadvantages are:
- No standard
- Nobody seems to use it, ergo no scrutiny and not battle-tested (hopefully we can change this)

balloon/src/lib.rs
daxpedda: Sure, there are still other things I will have to make sure are exactly the same, I will make an attempt.

tarcieri:
If we can generate test vectors and demonstrate some sort of interoperability between implementations in multiple languages, I think we could potentially consider including it.

Curious what @newpavlov thinks.

daxpedda:
> If we can generate test vectors and demonstrate some sort of interoperability between implementations in multiple languages, I think we could potentially consider including it.

I tested a couple, none of them are compatible with each other as far as I could determine:
- C prototype implementation: https://github.com/henrycg/balloon
- Rust: https://github.com/stichtingorganism/balloon
- Python: https://github.com/nachonavarro/balloon-hashing
- Go: https://github.com/nogoegst/balloon (not tested)
- C: https://github.com/deftchain/balloon_hash (not tested)

From what I could gather there are some minor differences in the implementation that keep things from being interoperable.

I currently started working on getting the prototype implementation into a Rust crate to test against: https://github.com/daxpedda/balloon-proto.
When working through the implementation, trying to figure out what the differences are exactly, I noticed that the prototype implementation significantly differs from the research paper. For example it hashes some things in a different order or adds completely new things to the hash. The mixing part of the the algorithm is the most confusing one, it really doesn't follow the research paper whatsoever, as far as I can tell.

Other implementations really followed the paper closely, I'm unsure what to do now, should I just try to reach compatibility with the prototype implementation, or should I try to follow the paper?

tarcieri:
Wow, that's not a great situation. I'm not sure what to do if literally no two implementations interoperate.

Perhaps it would be good to open some issues on the respective repos and try to coordinate some form of interoperability.

daxpedda:
I don't mind following up on that, but all of these implementations are years old by now, the prototype implementation is officially unsupported now: https://github.com/henrycg/balloon/issues/5#issuecomment-616425762. The recently active implementation is the Rust implementation: https://github.com/stichtingorganism/balloon.

newpavlov:
Lack of some kind of standardization, at the very least in the form of test vectors is certainly not great... I agree that it may be worth to try to coordinate some form of interoperability first.

I wonder if @henrycg can chime in and recommend whether we should treat the prototype code as a reference implementation. Also I wonder if @goldenMetteyya is open to sharing with us access to the `balloon` crate name on crates.io.

henrycg:
Yes, I am happy to add some context here, even if it is a bit unsatisfying.

For background: The Balloon algorithm never ended up being standardized in an RFC or similar standards document. The goal in the research paper was just to give a high-level description of the algorithm, without concern for the low-level details of encoding, endianness, ordering of arguments, etc.

Our hope at the time of writing the paper was that someone would be interested enough in the algorithm to formalize it in an RFC or similar standard at the level of precision that one would need to write an interoperable implementation. Unfortunately, that never happened. While the prototype code gives one way to map the high-level algorithm into code, there are lots of other ways that are equally good, or perhaps even better.

People who have used the algorithm in practice tend to make the low-level implementation decisions (about endianness, encoding, ordering of arguments, etc.) themselves. Since there is no standard implementation and since the algorithm is not widely used anyways, interoperability is not a concern for most users.

I'm not sure if this answers your question exactly, but I hope that the background is helpful.

daxpedda:
So I see three options:
1. We follow the paper closely, which would allow us to potentially be compatible with other libraries out there
2. We follow the prototype implementation, which would allow us to have some test vectors, but not really, because we will not use `EVP_EncryptUpdate` outside of testing
3. We close the PR, considering there is no standard or test vectors, it might have been a bad idea to begin this

I'm leaning towards option 1.

tarcieri:
Option 1 seems the best to me, but ideally producing a set of test vectors and ensuring interop with at least one other implementation, preferably implemented in something other than Rust

balloon/src/lib.rs
daxpedda: I changed `cnt` to be u64 to prevent overflow, but `i` can only reach 3, as big as `delta` basically. So protecting it from overflow won't be necessary. In any case, I had the brilliant idea to change everything into `u64`.

balloon/src/lib.rs
daxpedda: I had a branch working on it, but I guess our newest decision is to not try and mirror the prototype implementation, so I will close this. If we change our mind we can still revive it. The FFI crate is here: https://github.com/daxpedda/balloon-proto.

balloon/src/params.rs
daxpedda: The `Params` fields are already public, I made the `new` function to make it easier to create a `Params` struct, otherwise it would need `NonZeroU32::new(x).unwrap()` multiple times.

balloon/src/lib.rs
daxpedda: Done.

balloon/src/lib.rs
daxpedda: Done.

balloon/src/lib.rs
daxpedda: This is covered by converting it to `u64`.

balloon/src/error.rs
daxpedda: I added how much memory is needed in the documentation.

balloon/src/params.rs
daxpedda: So after some reading in the paper, I understood that compared to Argon2i, the space cost should be one quarter and the time cost should be the same. So I adjusted the defaults according to the defaults we set in Argon2.

Any other suggestions are welcome.

daxpedda:
I believe I addressed all remaining issues except the compatibility one. I will open a PR or issue at https://github.com/nachonavarro/balloon-hashing, which is the implementation with the most recent activity. If everything goes well I can add it to the tests.

balloon/src/lib.rs
daxpedda: For compatibility reasons I've revisited this.
Now that `crypto-bigint` supports modulus, I used it to properly apply modulus to the whole hash. A PR is missing to actually support this:
https://github.com/RustCrypto/crypto-bigint/pull/14

balloon/Cargo.toml
```diff
@@ -0,0 +1,34 @@
...
+documentation = "https://docs.rs/balloon-hash"
+repository = "https://github.com/RustCrypto/password-hashes/tree/master/balloon"
+keywords = ["crypto", "password", "hashing"]
+categories = ["cryptography", "no-std"]
+edition = "2018"
+readme = "README.md"
+
+[dependencies]
+digest = { version = "0.9", default-features = false }
+crypto-bigint = { version = "0.2", git = "https://github.com/RustCrypto/crypto-bigint", features = ["generic-array"] }
```
daxpedda, line 16: This needs an update now that https://github.com/RustCrypto/crypto-bigint/pull/14 is merged.
https://github.com/RustCrypto/crypto-bigint/pull/17 will allow us to use `default-features = false` here.

balloon/src/lib.rs
```diff
@@ -0,0 +1,395 @@
...
+/// - (Optional) Secret key a.k.a. "pepper" to be used
+#[derive(Clone, Default)]
+pub struct Balloon<'key, D: Digest>
+where
+    GenericArray<u8, D::OutputSize>: ArrayDecoding,
+    <GenericArray<u8, D::OutputSize> as ArrayDecoding>::Output:
+        Rem<NonZero<<GenericArray<u8, D::OutputSize> as ArrayDecoding>::Output>>,
+    <<GenericArray<u8, D::OutputSize> as ArrayDecoding>::Output as Rem<
+        NonZero<<GenericArray<u8, D::OutputSize> as ArrayDecoding>::Output>,
+    >>::Output: ArrayEncoding,
```
daxpedda, line 100: This is awful and can only be fixed if we further constrain `ArrayDecoding` and/or `ArrayEncoding` in crypto-bigint.

daxpedda: See https://github.com/RustCrypto/crypto-bigint/pull/18.

balloon/Cargo.toml
daxpedda: `default-features = false` was re-introduced.

balloon/src/lib.rs
daxpedda: RustCrypto/crypto-bigint#18 was merged removing these constraints.

balloon/Cargo.toml
daxpedda: All relevant PRs were merged, so this is only waiting for the next release.

tarcieri: Release PR here https://github.com/RustCrypto/crypto-bigint/pull/23

daxpedda: Thanks, done!

daxpedda: Done.

balloon/Cargo.toml
```diff
@@ -0,0 +1,35 @@
...
+[dependencies]
+digest = { version = "0.9", default-features = false }
+crypto-bigint = { version = "0.2.10", default-features = false, features = ["generic-array"] }
+
+# optional dependencies
+password-hash = { version = "0.3", default-features = false, optional = true }
+rayon = { version = "1", optional = true }
+
+[dev-dependencies]
+nachonavarro-balloon = { git = "https://github.com/khonsulabs/nachonavarro-balloon" }
```
daxpedda, line 23: Is having a git `dev-dependency` acceptable? Alternatives:
- Pin to a specific commit
- Release it on crates.io
- Move it's implementation inside of this crate

Ofc we are also willing to donate the crate to RustCrypto if it's preferred.

tarcieri: We can publish a crate with a `git` dev dependency, but the real downsides are it's outside our org and can break the build if it's in a bad state

daxpedda: As I said, we are willing to move the crate to the RustCrypto org, feel free to hit me up anytime for it.

daxpedda: I just added it right into the testing facilities, is that alright? It's easy to make it a separate crate, just say the word.

Also this adds a git submodule, I hope that's alright? If not, any advice on what to do instead would be appreciated. Just copying the file seemed problematic to me because of attribution and what not.

daxpedda: I also added automatic updates to the gitsubmodule via dependabot.

daxpedda: I agree, I'm unsure how to proceed though. What would you recommend?

> If not, any advice on what to do instead would be appreciated. Just copying the file seemed problematic to me because of attribution and what not.

rozbb:
Doesn't specifying the revision solve the problem of a breaking update? The only concern would be if upstream deleted the repo or did a force push in the future

tarcieri:
> The only concern would be if upstream deleted the repo or did a force push in the future

These things are still concerning

balloon/Cargo.toml
tarcieri: If it's OSS, and only for testing, I don't see an issue.

Vendor it into the repo and if need be I can work on the proper boilerplate for attribution.

newpavlov: I think it should be enough to simply generate a bunch of test vectors using `nachonavarro-balloon` and test against them instead of depending on it. We can simply leave a comment stating that the test vectors were generated using this library.

daxpedda: Done. I hope the test vectors already generated are enough, if not let me know!

daxpedda:
I am coordinating compatibility with a C implementation here: nogoegst/balloon#2.
Additionally, I made a PR adding test vectors to the Python implementation I provided compatibility to before: nachonavarro/balloon-hashing#3.

daxpedda:
Inspired by [nogoegst/balloon](https://github.com/nogoegst/balloon) and re-reading specific parts of the research paper, I decided to split the algorithm into two different variants: Balloon and Balloon-M. Specifically, Balloon doesn't support parallel computation, but Balloon-M does.

I wasn't entirely sure how to make sure parameters reflect the fact that only Balloon-M supports parallel computation, so I left it at documentation and returning errors. As always, any feedback would be appreciated.

Additionally I updated my PRs, nogoegst/balloon#2 and nachonavarro/balloon-hashing#3, to add test vectors for the Balloon-M variant. In the case of the Python implementation, I also added the Balloon-M variant to their implementation.

One last note, while working on compatibility between the Balloon-M variant of [nogoegst/balloon](https://github.com/nogoegst/balloon), I discovered that this implementation deviates a tiny bit from the research paper, in that the XOR'ed output of the parallel computed hashes was additionally hashed at the end. Upon some further research I discovered that XOR'ing hashes is generally considered cryptographically bad, therefore I added this approach in our implementation as well.

daxpedda:
[nachonavarro/balloon-hashing](https://github.com/nachonavarro/balloon-hashing) now runs all test vectors as part of their CI.

.github/workflows/workspace.yml
```diff
@@ -16,7 +16,7 @@ jobs:
       - uses: actions/checkout@v1
       - uses: actions-rs/toolchain@v1
         with:
-          toolchain: 1.51.0 # MSRV
+          toolchain: 1.56.1 # MSRV
```
daxpedda, line 19: I had to update the MSRV that is used by Clippy.

daxpedda:
Re-based and updated to digest and sha2 0.10.

.github/workflows/workspace.yml
daxpedda: EDIT: because of crypto-bigint, I don't remember why I said Clippy.

balloon/src/params.rs
```diff
@@ -0,0 +1,110 @@
...
+    core::convert::TryFrom,
+    password_hash::{errors::InvalidValue, ParamsString, PasswordHash},
+};
+
+/// Balloon password hash parameters.
+///
+/// These are parameters which can be encoded into a PHC hash string.
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub struct Params {
+    /// Space cost, expressed in of blocks.
```
newpavlov, line 16: "In number of blocks" or simply "in blocks"?

balloon/tests/balloon_m.rs
```diff
@@ -0,0 +1,102 @@
...
+        salt: b"salt",
+        s_cost: 1,
+        t_cost: 1,
+        p_cost: 1,
+        output: hex!("97a11df9382a788c781929831d409d3599e0b67ab452ef834718114efdcd1c6d"),
+    },
+];
+
+#[test]
+fn test() {
```
newpavlov, line 86: I have it's worth to have a more descriptive test name, e.g. `test_balloon_m`.

balloon/tests/balloon.rs
```diff
@@ -0,0 +1,72 @@
...
+        password: b"password",
+        salt: b"salt",
+        s_cost: 1,
+        t_cost: 1,
+        output: hex!("eefda4a8a75b461fa389c1dcfaf3e9dfacbc26f81f22e6f280d15cc18c417545"),
+    },
+];
+
+#[test]
+fn test() {
```
newpavlov, line 56: Same as with `tests/balloon_m.rs`. Rename to `test_balloon`?

balloon/src/lib.rs
```diff
@@ -0,0 +1,445 @@
...
+//! let parsed_hash = PasswordHash::new(&password_hash)?;
+//! assert!(balloon.verify_password(password, &parsed_hash).is_ok());
+//! # }
+//! # Ok(())
+//! # }
+//! ```
+//!
+//! [Balloon]: https://en.wikipedia.org/wiki/Balloon_hashing
+
+#![cfg_attr(not(feature = "std"), no_std)]
```
newpavlov, line 53: The most common way of handling optional `std`/`alloc` looks like this:
```rust
#![no_std]

#[cfg(feature = "alloc")]
extern crate alloc;

#[cfg(feature = "std")]
extern crate std;
```

balloon/tests/balloon_m.rs
```diff
@@ -0,0 +1,102 @@
+#![cfg(feature = "alloc")]
```
newpavlov, line 1: I think you can remove the `alloc` requirement by using `hash_with_memory` and allocating the necessary buffer inside of the test function. Test modules always have access to `std`.

newpavlov (commented):
I have several minor nits. I haven't looked deeply into implementation of the algorithm itself and I probably would've done some things a bit differently (e.g. I prefer free-standing functions, on top of which abstractions are built), but I think we can merge the crate after the nits will be fixed. Unless, of course, @tarcieri has something to add.

balloon/tests/balloon_m.rs
daxpedda: I named it `test_vectors`, it already is in a module called `balloon_m`, what do you think?

daxpedda:
> I prefer free-standing functions, on top of which abstractions are built

I could still do that, it sounds quiet attractive to me too.

daxpedda:
I hope I addressed all outstanding issues. Please feel free to ask for any kind of improvement, even if it's just a preference.

tarcieri:
Thank you!

ecki:
I am currently toying with a java impl, and i will try to make it compatible to this, thanks for the test vectors.

there are some open issues I tried to raise upstream, but maybe we can resolve it here?

* mod sampling bias of the bitstream
* Relative small size in sha256, go,sha512 or Blake?
* Pepper? Usage label
* Final (HMAC hashing).
* PBKDF2 preseeding (low itewtion instead of init seed?
* Salt guessing (underspecify salt so defender can half the tries
* delta configurable
* Xor mixing (more memory bw)
* Algorithm Id in modular password hash syntax

Still interested to look into this or has the argon fraction won? ,)

daxpedda:
IANAC but as I am generally interested in this algorithm here are my 2 cents:

> * Relative small size in sha256, go,sha512 or Blake?

The balloon-hash crate here already supports any hash with the `Digest` trait.
> * Pepper? Usage label

Also supported. Never heard of a usage label in a password hash.
> * Final (HMAC hashing).

Does this belong in a password hash?
> * delta configurable

Does that achieve anything?
> * Xor mixing (more memory bw)

I thought we are doing that already?
> * mod sampling bias of the bitstream
> * PBKDF2 preseeding (low itewtion instead of init seed?
> * Salt guessing (underspecify salt so defender can half the tries
> * Algorithm Id in modular password hash syntax

No clue what these are.
> Still interested to look into this or has the argon fraction won? ,)

Honestly if Balloon is as good as the research says it is I think it's sad that it was forgotten like this.
The only way I see Balloon can be salvaged is if somebody makes a serious standardization effort, at the IETF for example.

ecki:
> > * Relative small size in sha256, go,sha512 or Blake?
> 
> The balloon-hash crate here already supports any hash with the `Digest` trait.

yes, But there is no reference impl and no (official?) test vectors, so it would be good if „community“ can agree on one alternate hash which is supported by all and does not have the blocksize problem of the AB16b paper (bw is worse than argon2).

> > * Pepper? Usage label
> 
> Also supported. Never heard of a usage label in a password hash.

if you use it for KDF it might be relevant. Not sure how safe the idea of using it as a KDF would be though. For password hashes a „site distinguisher“ would be a Pepper with lower secrecy requirements. It would prevent reusing „found hashes“ (known from the md5 correlation attacks (shucking)). Not sure how important that is for modern/salted hashes, but still it does not cost much. Anyway, I will implement a pepper interface I balloon compatible to your rust api - if I can find it ,)

> > * Final (HMAC hashing).
> 
> Does this belong in a password hash?

the balloon paper suggests a hash or HMAC as a final step. It is just a small extension after the xor step (or to replace the xor step). If you use a real HMAC in this step it would introduce a good place for a „NIST compliant“ peppering step.

> > * delta configurable
> 
> Does that achieve anything?

it would add confusion, but more seriously it looks like it was a moving target in the beginning and the xx paper identified a much larger delta as a possibility to gain some protection back (in their proposed xor alternative). 

> > * Xor mixing (more memory bw)
> 
> I thought we are doing that already?

Yes the parallel phase uses it already, but I meant in the mixing phase as discussed by AB16b. (But i know it comes with its own problems as the xor-block states could be used as a optimized way of storage. So that would require some more advanced crypto knowledge to design it it, I am curious if there is some interest for follow up research)

> > * mod sampling bias of the bitstream

currently the mix function retrieves a 64Bit pseudorandom integer and reduces it to the neighbor blockindex with integer module, this introduces some bias as the block numbers are not power of two. This hohle be fixed but would require a new algorithm version. I also wonder if 32bit for Block Index would be enough.

> > * PBKDF2 preseeding (with lower iterations) instead of init seed?

when hashing password and seed into the initial state with the HMAC iterations of PBKDF2 you have nearly no increase in complexity but an „approved“ function which would „firewall“ password from the relative uncommon and under-analysed balloon function.

> > * Salt guessing (underspecify salt so defender can half the tries

It is a general trick to „forget“ some parts of the salt or randomize the iteration numbers. This forces the validator to try all combinations (for example 4 different 2 bit combinations of the salt). As a defender you normally can validate a correct password after half the attempts by chance, as a attacker you would have to try all 4 wrong attempts - effectively halving the attackers cracking advantage (with the cost of making the verification times a bit more unpredictable if not run in parallel)

> > * Algorithm Id in modular password hash syntax

we should see which hash, which algorithm modifications, maybe which bitstream method, pre/posthashinng and random-module is used. For example we can define `$balloon$v=1$a=0..` parameter (add it to the meta dat hashing if >0). 0 would be sha256 with aes-ctr-sha256, a=1 could be sha512 etc. (the bitstream fix could also be v=2).

another parameter would be good to name the pepper key for rotation, but that one most likely does not need standardization.

> Honestly if Balloon is as good as the research says it is ..

i also like its construction simplicity, but regarding „as good as research“, there is not much (and it’s really weird NIST recommended it even without any production ready implementwtion) did you check the alwen,blocki 16  paper?

AB16b: https://eprint.iacr.org/2016/759

BTW if you or any reader/researcher also are on the balloon-hashing mailinglist, should we continue the discussion there?

tarcieri:
> BTW if you or any reader/researcher also are on the balloon-hashing mailinglist

TIL there's a mailing list. I guess it's this one? https://mailman.stanford.edu/mailman/listinfo/balloon-hashing

daxpedda:
@ecki honestly most of this stuff goes way over my head, as I said, I am not a cryptographer.

I am happy to discuss or adapt the implementation, or add or provide more test vectors if you have any specific suggestions.

My only strong opinion on this whole matter is that if we really want to improve the current situation, some cryptographer(s) will have to get together and push for standardization at the IETF, like Argon2 or anything else really.

samuel-lucas6:
Hi folks, I'm writing an [Internet-Draft](https://github.com/samuel-lucas6/draft-lucas-balloon-hashing) for Balloon so there's an actual specification and would like to ask for feedback on the current draft and input on the open [Issues](https://github.com/samuel-lucas6/draft-lucas-balloon-hashing/issues). Some ideas are breaking changes but would improve the algorithm. Another important aspect is recommending parameters, which requires proper benchmarking like what Steve Thomas has [done](https://tobtu.com/minimum-password-settings/). Any help would be greatly appreciated.

I reached out to Henry Corrigan-Gibbs, but he's busy with teaching/research commitments. I tried emailing some people on the PHC panel, but nobody replied. A few people don't have a public email address, and I thought there was no point emailing you @tarcieri when leaving this message here. I haven't yet tried contacting the authors of the papers looking at Balloon, but I'm not sure how successful that will be either given they are a few years old now and the algorithm has been modified.

I'm not expecting the Internet-Draft to become an RFC given Argon2's success, but the algorithm could still see some usage like XChaCha20.

ecki:
@samuel-lucas6 i would like an I-D, especially given NISTs premature recommendation of the algorithm. A incompatible variant would allow to fix some of the issues, so it’s a good starting point:
- require UTF-8 encoding of password bytes and pre-hashing
- Specify optional salt opportunity
- Fix the (small) modulo bias for the RNG
- Define a standard modular hash syntax (including a seed Id)

do you also plan to address the scheduling weakness found in the follow-up paper? (This night however require some academic involvement?)

samuel-lucas6:
@ecki

> require UTF-8 encoding of password bytes and pre-hashing

Not sure what you mean by pre-hashing here. I originally specified that the password should be UTF-8 encoded, but I noticed that the Argon2 and scrypt RFCs don't mention this so removed it.

> Specify optional pepper opportunity (secret key)

See [this](https://github.com/samuel-lucas6/draft-lucas-balloon-hashing/issues/7) issue.

> Fix the (small) modulo bias for the RNG

This hasn't been addressed. The current draft is compatible with existing interoperable implementations.

> Define a standard modular hash syntax (including a pepper Id)

This has been [attempted](https://samuellucas.com/draft-lucas-balloon-hashing/draft-lucas-balloon-hashing.html#name-encoding-password-hashes) minus the pepper part, but it's quite messy because of Balloon vs Balloon-M. I would prefer if there was just one algorithm, as mentioned [here](https://github.com/samuel-lucas6/draft-lucas-balloon-hashing/issues/5).

> do you also plan to address the scheduling weakness found in the follow-up paper? (This night however require some academic involvement?)

I haven't listed that down as I haven't tried to read all the third-party papers yet. That may not be feasible as the designers have moved on. I was planning on making minor design modifications at most as I'm not a cryptographer and no cryptographers seem to be that interested in Balloon.

Issue #109, opened 2021-01-29
argon2: test against full set of vectors from reference implementation


tarcieri:
Presently the `argon2` crate tests include Known Answer Tests (KAT) and some PHC string tests.

However, the [`test.c`](https://github.com/P-H-C/phc-winner-argon2/blob/master/src/test.c) file from the Argon2 reference implementation includes a large number of test vectors which we don't presently test against.

We should incorporate all of these test vectors into the `argon2` crate's test suite.

PR #249, opened 2021-10-13
argon2: Add tests from reference implementation

touilleMan:
Hi !

I tried to add the tests from the argon2 reference implementation as suggested in #109

It's not finished, but I'd like to have some pointers there 😃 

Most (but not all !) tests involving argon2i are falling on `ctx.verify_password` (but `hash_password_into` returns the correct hash).
On the other hand argon2id tests are passing, so I guess the issue comes from the way I use verify_password...

On top of that I'm not sure how to implement the test related to the encoding errors (see commented part of the code):
In the reference implementation, test are done to check the hash function detect incorrect hash encoding. However this doesn't work given `PasswordHash::new` does the parsing and is ok with missing fields.

See for instance the `reference_argon2i_invalid_encoding_1` test that provide an encoded hash with a missing $, we end up with:
```
PasswordHash { algorithm: Ident("argon2i"), version: None, params: {Ident("m"): Value("65536"), Ident("t"): Value("2"), Ident("p"): Value("1c29tZXNhbHQ")}, salt: Some(Salt("9sTbSlTio3Biev89thdrlKKiCaYsjjYVJxGAL3swxpQ")), hash: None }
```
But argon doesn't seem to complain (even from the `p=1c29tZXNhbHQ` !)

I also don't understand how to use the t/p/m configuration coming from the encoded hash: it seems incredibly cumbersome to have to use `ParamsBuilder` in conjunction with the `PasswordHash` object to manually extract and validate parameters...

anyway I waiting for your feedback ;-)

### DIFF
--- a/argon2/tests/kat.rs
+++ b/argon2/tests/kat.rs
@@ -402,9 +402,31 @@ macro_rules! testcase_good {
     };
 }
 
+macro_rules! ignored_testcase_good {
+    ($name: ident, $algorithm: expr, $version: expr, $t: expr, $m: expr, $p: expr, $pwd: expr, $salt: expr, $expected_raw: expr, $expected_phc: expr, $alternative_phc: expr) => {
+        #[test]
+        #[ignore]
+        fn $name() {
+            hashtest(
+                $algorithm,
+                $version,
+                $t,
+                $m,
+                $p,
+                $pwd,
+                $salt,
+                $expected_raw,
+                $expected_phc,
+                $alternative_phc,
+            )
+        }
+    };
+}
+
 /* Argon2i V0x10: Multiple test cases for various input values */
 
-testcase_good!(
+// TODO: If version is not provided, verifier incorrectly uses version 0x13
+ignored_testcase_good!(
     reference_argon2i_v0x10_2_16_1,
     Algorithm::Argon2i,
     Version::V0x10,
@@ -418,6 +440,8 @@ testcase_good!(
     "$argon2i$m=65536,t=2,p=1$c29tZXNhbHQ$9sTbSlTio3Biev89thdrlKKiCaYsjjYVJxGAL3swxpQ"
 );
 
+// TODO: If version is not provided, verifier incorrectly uses version 0x13
+#[ignore]
 #[cfg(feature = "test_large_ram")]
 testcase_good!(
     reference_argon2i_v0x10_2_20_1_large_ram,
@@ -433,7 +457,8 @@ testcase_good!(
     "$argon2i$m=1048576,t=2,p=1$c29tZXNhbHQ$lpDsVdKNPtMlYvLnPqYrArAYdXZDoq5ueVKEWd6BBuk"
 );
 
-testcase_good!(
+// TODO: If version is not provided, verifier incorrectly uses version 0x13
+ignored_testcase_good!(
     reference_argon2i_v0x10_2_18_1,
     Algorithm::Argon2i,
     Version::V0x10,
@@ -447,7 +472,8 @@ testcase_good!(
     "$argon2i$m=262144,t=2,p=1$c29tZXNhbHQ$Pmiaqj0op3zyvHKlGsUxZnYXURgvHuKS4/Z3p9pMJGc"
 );
 
-testcase_good!(
+// TODO: If version is not provided, verifier incorrectly uses version 0x13
+ignored_testcase_good!(
     reference_argon2i_v0x10_2_8_1,
     Algorithm::Argon2i,
     Version::V0x10,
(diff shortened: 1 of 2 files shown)


argon2/tests/kat.rs
```diff
@@ -288,3 +290,572 @@ fn argon2id_v0x13() {
...
+
+/* Argon2i V0x10: Error state tests */
+
+#[test]
+fn reference_argon2i_invalid_encoding_1() {
+    /* Handle an invalid encoding correctly (it is missing a $) */
+    let hash = PasswordHash::new(
+        "$argon2i$m=65536,t=2,p=1c29tZXNhbHQ$9sTbSlTio3Biev89thdrlKKiCaYsjjYVJxGAL3swxpQ",
+    )
+    .unwrap();
```
tarcieri, line 525: A quick note that this isn't expected to immediately fail re: your comment here:

> But argon doesn't seem to complain (even from the p=1c29tZXNhbHQ !)

`PasswordHash` comes from the `password-hash` crate and doesn't eagerly validate the parameters for a particular algorithm:

https://docs.rs/password-hash/0.3.2/password_hash/struct.PasswordHash.html

The parameters can be validated by extracting them using the `TryFrom<&PasswordHash>` impl on `Params`: 

https://docs.rs/argon2/0.3.1/argon2/struct.Params.html#impl-TryFrom%3C%26%27a%20PasswordHash%3C%27a%3E%3E

touilleMan:
> I'm a bit unclear what the problem you're describing here is. You shouldn't need to use ParamsBuilder to "extract and validate parameters" at all, only build them.
>
> Instead you should use the Params struct, which is intend to always represent valid parameters: https://docs.rs/argon2/0.3.1/argon2/struct.Params.html

Thanks a lot for this feedback, I finally figured out how I was supposed to do it 😃 
```rust
let hash = PasswordHash::new(
    "$argon2i$m=65536,t=2,p=1c29tZXNhbHQ$9sTbSlTio3Biev89thdrlKKiCaYsjjYVJxGAL3swxpQ",
)
.unwrap();
let ret = Params::try_from(&hash);
assert_eq!(ret, Err(Error::ParamValueInvalid(InvalidValue::InvalidChar('c'))));
```

As a new user, I understood wrong the [example on the doc](https://docs.rs/argon2/0.3.1/argon2/index.html#usage-simple-with-default-params):
```rust
// Argon2 with default params (Argon2id v19)
let argon2 = Argon2::default();
[...]
// Verify password against PHC string
let parsed_hash = PasswordHash::new(&password_hash)?;
assert!(argon2.verify_password(password, &parsed_hash).is_ok());
```

I looked like the `verify_password` was configured by the `Argon2::default()` call (given `Argon2` contains the algo type and t/m/p stuff). Hence my comment about manually extracting parameters from the password hash to build the correct `Argon2` instance then use it to call `verify_password` :/

From this on, I dig a lot into `argon2` and `password_hash` crates, it took me many hours and you message to understand what I missed 😄 .

So I think it would be better to provide `verify_password` as a regular function (typically calling `Argon2::default().verify_password` under the hood), would you accept a PR on this ?

Ideally `Argon2::verify_password` (and `hash_password_customized`) should be an associated functions, then `PasswordVerifier` could also be modified:
```rust
impl<T: PasswordHasher> PasswordVerifier for T {
    fn verify_password(password: &[u8], hash: &PasswordHash<'_>) -> Result<()> {
        if let (Some(salt), Some(expected_output)) = (&hash.salt, &hash.hash) {
            let computed_hash = T::hash_password_customized(
                password,
                Some(hash.algorithm),
                hash.version,
                T::Params::try_from(hash)?,
                *salt,
            )?;

            if let Some(computed_output) = &computed_hash.hash {
                // See notes on `Output` about the use of a constant-time comparison
                if expected_output == computed_output {
                    return Ok(());
                }
            }
        }

        Err(Error::Password)
    }
}
```
Of course this would break compatibility, but I think it would be a great improvement for the public API (and internal code readability, especially given finding where `verify_password` is defined is also challenging ! 😄 )

argon2/src/lib.rs
```diff
@@ -53,6 +53,7 @@
 //!
 //! // Verify password against PHC string
 //! let parsed_hash = PasswordHash::new(&password_hash)?;
+//! Note hash params from `parsed_hash` are used instead of what is configured in `argon2` here !
```
tarcieri, line 56: This is failing to parse because it needs to be a nested comment.

```suggestion
//! // Note hash params from `parsed_hash` are used instead of what is configured in `argon2` here !
```

tarcieri:
@touilleMan good progress! Looks like the doctests are failing.

touilleMan:
@tarcieri I've updated the last commit to correct the documentation

I've also modified the example to use `Argon2::default().verify_password`. This is to isolate the hash and verify code snippets (given in real life you never use for verifying a hash the very same `argon2` instance that have been used to hash it in the first place ^^)

tarcieri:
Looks good now! Thanks for contributing these!

touilleMan:
Ha ! you went too fast !

I made a typo in the doc:
```diff
@@ -53,7 +53,7 @@
 //!
 //! // Verify password against PHC string
 //! let parsed_hash = PasswordHash::new(&password_hash)?;
-//! // Note hash params from `parsed_hash` are used instead of what is configured in the `Argron2` instance
+//! // Note hash params from `parsed_hash` are used instead of what is configured in the `Argon2` instance
```

Can you fix it on master or do you need another PR ?

PR #255, opened 2021-11-24
ability to use custom output key length for scrypt

Numenorean:
I've faced with an issue when i need dynamically specify the output length of scrypt function. With current state of your crate it's not possible because in `Params::new` there is a statically used output hash length (RECOMMENDED_LEN = 64 bytes), so i don't understand why it's "recommended" if there is no choice

### DIFF
--- a/scrypt/src/params.rs
+++ b/scrypt/src/params.rs
@@ -77,7 +77,7 @@ impl Params {
             log_n,
             r: r as u32,
             p: p as u32,
-            len: len,
+            len,
         })
     }
 
--- a/scrypt/src/params.rs
+++ b/scrypt/src/params.rs
@@ -29,15 +29,18 @@ impl Params {
     /// - `log_n` - The log2 of the Scrypt parameter `N`
     /// - `r` - The Scrypt parameter `r`
     /// - `p` - The Scrypt parameter `p`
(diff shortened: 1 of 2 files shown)


newpavlov (commented):
Looks good. The only small concern I have is that `len` is used only in the `PasswordHasher` impl, while the `scrypt` function does not touch this field at all.

Since this is a breaking change, I think we will release it after the changes from #254.

I will leave this PR open for now in the case if @tarcieri has anything to add. Meanwhile can you please fix the clippy warning, either by applying or disabling  the lint?

tarcieri:
Looks fine to me. `argon2::Params` works somewhat similarly:

https://docs.rs/argon2/latest/argon2/struct.Params.html

The main difference is it accepts an `Option` for the output size, however the logic isn't currently in place to handle that, and it could be fixed in a followup.

PineappleIOnic:
@tarcieri What is the status of this PR? Is there any chance it could get merged?

tarcieri:
CI is failing due to a clippy error. If someone would like to pick it up again and get CI green, we can merge it.

tarcieri:
Will fix the clippy error retroactively.

This is a breaking change so I'll also bump to a `-pre`.

PR #258, opened 2021-12-21
Make `Params::block_count()` public

### LINKED ISSUE #257: argon2: Missing access to `Params::block_count()`
It's currently a bit difficult to use `Argon2::hash_password_into_with_memory` without knowing exactly how much memory is required. Can this function just be made public?

daxpedda:
Fixes #257.


tarcieri:
Thanks!

PR #277, opened 2022-02-20
[bcrypt-pbkdf] fixes #276 -- allow passphrase to be impl AsRef<[u8]> to permit non-utf8 passphrases

### LINKED ISSUE #276: bcrypt_pbkdf: Support passphase that's &[u8] and not &str
The only thing that's done with the passphrase is pass `Sha512::digest(passphrase.as_bytes())` so it should be fine to allow it to be arbitrary bytes.

This probably needs to be done by either a) making passphrase's type generic, b) adding a second function

alex:


alex:
@tarcieri Assuming this is acceptable, would you be willing to do a bcrypt-pbkdf release? This appears to be the last missing piece before I can propose migrating https://github.com/pyca/bcrypt to Rust :-) Happy to do a PR with the version bump + changelog.

tarcieri (approved):

tarcieri:
@alex sure, sounds good

PR #278, opened 2022-02-20
bcrypt-pbkdf: bump for 0.8.1 release

alex:


alex:
I went with 0.8.1 since it's a backwards compatible change and so didn't justify a semver bump. Let me know if there's a preference making it 0.9 instead, since it is a new feature.

PR #306, opened 2022-06-08
Argon2: add constant for recommended salt length

falko17:
The specification of argon2 recommends a salt length of 16 bytes for password hashing (section 3.1).
This pull request introduces a constant `RECOMMENDED_SALT_LEN` to make this recommended salt length accessible to users.
(Also, a slight mistake in the documentation has been corrected.)

### DIFF
--- a/argon2/src/lib.rs
+++ b/argon2/src/lib.rs
@@ -106,6 +106,11 @@ pub const MIN_SALT_LEN: usize = 8;
 /// Maximum salt length in bytes.
 pub const MAX_SALT_LEN: usize = 0xFFFFFFFF;
 
+/// Recommended salt length for password hashing in bytes.
+#[cfg(feature = "password-hash")]
+#[cfg_attr(docsrs, doc(cfg(feature = "password-hash")))]
+pub const RECOMMENDED_SALT_LEN: usize = password_hash::Salt::RECOMMENDED_LENGTH;
+
 /// Maximum secret key length in bytes.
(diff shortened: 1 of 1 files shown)


falko17:
Oh, that's true, didn't notice that. Should I remove `RECOMMENDED_SALT_LEN` again, or should I just set its value to `Salt::RECOMMENDED_LENGTH`?

tarcieri:
The latter (effectively re-exporting the constant) is fine.

falko17:
Sorry for the notifications, I was stumbling around with the features a bit. I believe it should work now, though the new constant is only available if the `password-hash` feature is enabled.

argon2/src/lib.rs
```diff
@@ -100,12 +100,16 @@ use password_hash::{Decimal, Ident, ParamsString, Salt};
...
 
-/// Minimum and maximum salt length in bytes.
+/// Minimum salt length in bytes.
 pub const MIN_SALT_LEN: usize = 8;
 
 /// Maximum salt length in bytes.
 pub const MAX_SALT_LEN: usize = 0xFFFFFFFF;
 
+/// Recommended salt length for password hashing in bytes.
+#[cfg(all(feature = "password-hash"))]
```
tarcieri, line 110:
```suggestion
#[cfg(feature = "password-hash")]
#[cfg_attr(docsrs, doc(cfg(feature = "password-hash")))]
```

falko17: Applied in d298106586d3cf20b0e24372e0fcb600f15c5987.

PR #307, opened 2022-06-10
Remove crate feature from `argon2::RECOMMENDED_SALT_LEN`

daxpedda:
A small follow-up to https://github.com/RustCrypto/password-hashes/pull/306.

The proposition here is to be able to use `RECOMMENDED_SALT_LEN` without having to pull in another dependency. Is that alright?

Sorry for being so brusk, I thought instead of asking with an issue I can simply ask with a PR.


tarcieri:
Yes, this is fine

Issue #310, opened 2022-06-15
balloon-hash: provide `hash_password_into` functionality


brxken128:
I'm in the process of migrating from `argon2id` to balloon hashing, but I'm unable to do so in a relatively-safe manner.

I relied on `argon2`'s `hash_password_into` in order to keep track of where the key is, and prevent potential cloning and such. `balloon-hash` doesn't seem to provide this functionality, and instead returns a GenericArray - which complicates this task further.

What should've been a few lines now requires lots of manual zeroizing, and I just don't think it's a great alternative.

Would it be possible to provide `hash_password_into` functionality? I had a quick peek into the `argon2` crate and it doesn't seem out of scope, but I'm not familiar with the RustCrypto traits, so I don't think it's best for me to attempt.

Thank you.

PR #311, opened 2022-06-16
Zeroize sensitive bytes for `balloon-hash`

daxpedda:
Zeroizing the hash output when cloning.

~~I realized by #310 that we don't really zeroize some leftover bytes. Looking at `argon2` I noticed that zeroizing bytes is behind a crate feature, does that make sense? Considering this is relevant for security, it seems to me making this optional wouldn't make sense.~~

~~Other questions that arose:~~
- ~~Currently a lot of types being returned here don't zeroize on drop. For example, using `hash` returns a `GenericArray`. I was thinking of introducing a type that implements `ZeroizeOnDrop`, but realized maybe this type should be shared among a lot of crates. Should we create a tracking issue and discuss this or should I implement something rudimentary in `balloon-hash` for now?~~
- ~~I noticed that `Digest` doesn't do any zeroizing at all. As far as I understand, the hash state shouldn't be leaked either and all of this should be zeroized when everything's done. Again, is a tracking issue for this appropriate?~~

EDIT: Resolved some questions over Zulip, TLDR, I'm not a cryptographer and what I was intending was most likely overkill.

### DIFF
--- a/balloon-hash/src/balloon.rs
+++ b/balloon-hash/src/balloon.rs
@@ -137,7 +137,7 @@ where
     // for m from 1 to s_cost-1:
     for m in 1..s_cost {
         // buf[m] = hash(cnt++, buf[m-1])
-        Digest::update(&mut digest, &cnt.to_le_bytes());
+        Digest::update(&mut digest, cnt.to_le_bytes());
         cnt += 1;
         Digest::update(&mut digest, &buf[m - 1]);
         buf[m] = digest.finalize_reset();
@@ -157,7 +157,7 @@ where
             };
 
             // buf[m] = hash(cnt++, prev, buf[m])
-            Digest::update(&mut digest, &cnt.to_le_bytes());
+            Digest::update(&mut digest, cnt.to_le_bytes());
             cnt += 1;
             Digest::update(&mut digest, prev);
             Digest::update(&mut digest, &buf[m]);
@@ -167,13 +167,13 @@ where
             // for i from 0 to delta-1:
             for i in 0..DELTA {
                 // block_t idx_block = ints_to_block(t, m, i)
-                Digest::update(&mut digest, &t.to_le_bytes());
-                Digest::update(&mut digest, &(m as u64).to_le_bytes());
-                Digest::update(&mut digest, &i.to_le_bytes());
+                Digest::update(&mut digest, t.to_le_bytes());
+                Digest::update(&mut digest, (m as u64).to_le_bytes());
+                Digest::update(&mut digest, i.to_le_bytes());
                 let idx_block = digest.finalize_reset();
 
                 // int other = to_int(hash(cnt++, salt, idx_block)) mod s_cost
-                Digest::update(&mut digest, &cnt.to_le_bytes());
+                Digest::update(&mut digest, cnt.to_le_bytes());
                 cnt += 1;
                 Digest::update(&mut digest, salt);
 
@@ -194,7 +194,7 @@ where
                 );
 
                 // buf[m] = hash(cnt++, buf[m], buf[other])
-                Digest::update(&mut digest, &cnt.to_le_bytes());
+                Digest::update(&mut digest, cnt.to_le_bytes());
                 cnt += 1;
                 Digest::update(&mut digest, &buf[m]);
                 Digest::update(&mut digest, &buf[other]);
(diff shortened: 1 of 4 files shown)


balloon-hash/src/balloon.rs
```diff
@@ -179,25 +185,30 @@ where
                     Digest::update(&mut digest, thread_id.to_le_bytes());
                 }
 
-                Digest::update(&mut digest, idx_block);
+                Digest::update(&mut digest, &idx_block);
+                idx_block.zeroize();
```
daxpedda, line 189: Does this make sense?

daxpedda: After some discussion on Zulip I decided to only zeroize things that are really necessary. Like the hash output.

balloon-hash/src/balloon.rs
```diff
@@ -179,25 +185,30 @@ where
...
                         .unwrap(),
                 );
 
                 // buf[m] = hash(cnt++, buf[m], buf[other])
-                Digest::update(&mut digest, &cnt.to_le_bytes());
+                Digest::update(&mut digest, cnt.to_le_bytes());
                 cnt += 1;
                 Digest::update(&mut digest, &buf[m]);
                 Digest::update(&mut digest, &buf[other]);
+                other.zeroize();
```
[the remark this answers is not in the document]
daxpedda: See above.

balloon-hash/src/balloon.rs
```diff
@@ -34,11 +37,15 @@ where
...
-
-        for thread in 1..=u64::from(params.p_cost.get()) {
-            let hash = hash_internal::<D>(pwd, salt, secret, params, memory_blocks, Some(thread))?;
-            output = output.into_iter().zip(hash).map(|(a, b)| a ^ b).collect();
+        let mut output = hash_internal::<D>(pwd, salt, secret, params, memory_blocks, Some(1))?;
+
+        for thread in 2..=u64::from(params.p_cost.get()) {
+            #[cfg_attr(not(feature = "zeroize"), allow(unused_mut))]
+            let mut hash =
+                hash_internal::<D>(pwd, salt, secret, params, memory_blocks, Some(thread))?;
```
newpavlov, line 45: You can move `let mut hash` outside of the loop and write something like this:
```rust
let mut hash = Default::default();
for thread in 2..=u64::from(params.p_cost.get()) {
    hash = ...;
    // ...
}
hash.zeroize();
```

balloon-hash/src/balloon.rs
```diff
@@ -62,9 +69,16 @@ where
...
-                    Ok(a.into_iter().zip(b).map(|(a, b)| a ^ b).collect())
-                })
+                .try_reduce(
+                    GenericArray::default,
+                    |mut a, #[cfg_attr(not(feature = "zeroize"), allow(unused_mut))] mut b| {
+                        a.iter_mut().zip(&b).for_each(|(a, b)| *a ^= b);
+                        #[cfg(feature = "zeroize")]
+                        b.zeroize();
+
+                        Ok(a)
```
newpavlov, line 79: Here a better solution could look like this:
```rust
|mut a, b| {
    a.iter_mut().zip(&b).for_each(|(a, b)| *a ^= b);
    #[cfg(feature = "zeroize")]
    Zeroizing::new(b)
    Ok(a)
}
```

newpavlov (commented):
Honestly, I am a bit skeptical about zeroizing values on stack. It's hard to make it reliable and there are often a bunch of copies of "zeroized" data which stay on stack. For stack data we should use stack bleaching, not zeroizing of temporary stack-based variables.

balloon-hash/src/balloon.rs
daxpedda: :+1:
Is there something I should know about why this is better except skipping the whole `cfg` because of `mut`, or is that it?

newpavlov: No, the `cfg_attr` simply looked a bit too noisy to me.

daxpedda:
> Honestly, I am a bit skeptical about zeroizing values on stack. It's hard to make it reliable and there are often a bunch of copies of "zeroized" data which stay on stack. For stack data we should use stack bleaching after sensitive function was executed, not zeroizing of temporary stack-based variables.

I read about this before, but as far as I know, this isn't really a thing for Rust right now, is that correct?
Should I remove all stack data zeroizing because it's completely pointless? Or is it better then nothing?

balloon-hash/src/balloon.rs
daxpedda: Sadly `GenericArray` can't be wrapped into `Zeroizing` :(.

balloon-hash/src/balloon.rs
daxpedda: Done.

daxpedda:
See #312.

PR #313, opened 2022-06-16
Implement `Balloon::hash_into`

### LINKED ISSUE #310: balloon-hash: provide `hash_password_into` functionality

daxpedda:
~~Depends on #312.~~
Fixes #310.

### DIFF
--- a/balloon-hash/src/error.rs
+++ b/balloon-hash/src/error.rs
@@ -3,7 +3,7 @@
 use core::fmt;
 
 #[cfg(feature = "password-hash")]
-use ::{core::cmp::Ordering, password_hash::errors::InvalidValue};
+use {core::cmp::Ordering, password_hash::errors::InvalidValue};
 
 /// Result with balloon's [`Error`] type.
 pub type Result<T> = core::result::Result<T, Error>;
@@ -56,6 +56,7 @@ impl From<Error> for password_hash::Error {
             Error::ThreadsTooFew => InvalidValue::TooShort.param_error(),
             Error::ThreadsTooMany => InvalidValue::TooLong.param_error(),
             Error::TimeTooSmall => InvalidValue::TooShort.param_error(),
+            // TODO: Update after RustCrypto/traits#1026.
             Error::OutputSize { actual, expected } => match actual.cmp(&expected) {
                 Ordering::Less => password_hash::Error::OutputTooShort,
                 Ordering::Greater => password_hash::Error::OutputTooLong,
--- a/balloon-hash/src/error.rs
+++ b/balloon-hash/src/error.rs
@@ -3,7 +3,7 @@
 use core::fmt;
 
 #[cfg(feature = "password-hash")]
-use password_hash::errors::InvalidValue;
+use ::{core::cmp::Ordering, password_hash::errors::InvalidValue};
 
 /// Result with balloon's [`Error`] type.
 pub type Result<T> = core::result::Result<T, Error>;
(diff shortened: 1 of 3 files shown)


balloon-hash/src/error.rs
```diff
@@ -21,6 +21,8 @@ pub enum Error {
     ThreadsTooMany,
     /// Time cost is too small.
     TimeTooSmall,
+    /// Output size not correct.
+    OutputSize,
```
newpavlov, line 25: Note that adding this error variant makes this PR a breaking one.

balloon-hash/src/error.rs
```diff
@@ -45,6 +48,7 @@ impl From<Error> for password_hash::Error {
             Error::ThreadsTooFew => InvalidValue::TooShort.param_error(),
             Error::ThreadsTooMany => InvalidValue::TooLong.param_error(),
             Error::TimeTooSmall => InvalidValue::TooShort.param_error(),
+            Error::OutputSize => password_hash::Error::OutputTooShort,
```
daxpedda, line 51: This isn't exactly correct, but I hope it's alright? Alternatively we add a new error variant to `password_hash::Error`?

balloon-hash/src/error.rs
daxpedda: Yeah, I don't really mind if this can't be merged until the next release. I'm not really sure what else to do here, alternatively of course I can return a wrong error type, just to make it a non-breaking change.

tarcieri: This crate doesn’t have many users so you could just go ahead and make a breaking change.

If you also add `#[non_exhaustive]` to the error enum you can make such changes non-breaking going forward.

daxpedda: Done.

balloon-hash/src/error.rs
tarcieri: Can you open an issue on https://github.com/rustcrypto/traits ?

daxpedda: I just opened a PR instead https://github.com/RustCrypto/traits/pull/1026, hope that's okay.

tarcieri: Sure, except it seems like if you changed `balloon_hash::Error::OutputSize` to track an expected/actual value similar to what you propose in https://github.com/RustCrypto/traits/pull/1026 it should be possible to conditionally emit `password_hash::Error::OutputTooShort`/`OutputTooLong` correctly

daxpedda: Yeah, that's a good idea.

daxpedda: Done.

balloon-hash/src/error.rs
```diff
@@ -3,13 +3,14 @@
 use core::fmt;
 
 #[cfg(feature = "password-hash")]
-use password_hash::errors::InvalidValue;
+use ::{core::cmp::Ordering, password_hash::errors::InvalidValue};
```
tarcieri, line 6: The `::` here is unnecessary.

```suggestion
use {core::cmp::Ordering, password_hash::errors::InvalidValue};
```

daxpedda: Didn't know that :astonished:.
Done.

balloon-hash/src/error.rs
```diff
@@ -45,6 +56,11 @@ impl From<Error> for password_hash::Error {
             Error::ThreadsTooFew => InvalidValue::TooShort.param_error(),
             Error::ThreadsTooMany => InvalidValue::TooLong.param_error(),
             Error::TimeTooSmall => InvalidValue::TooShort.param_error(),
+            Error::OutputSize { actual, expected } => match actual.cmp(&expected) {
+                Ordering::Less => password_hash::Error::OutputTooShort,
+                Ordering::Greater => password_hash::Error::OutputTooLong,
+                Ordering::Equal => unreachable!("unexpected correct output size"),
```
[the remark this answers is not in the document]
daxpedda: Done.

tarcieri:
Thanks!

PR #324, opened 2022-07-12
doc: Add a link to scrypt's usage example

nathanielknight:
I ended up on the [`Scrypt` struct's page] while trying to figure out how to use the crate and it wasn't obvious what the associated function to actually hash a password was. This commit adds a link to the crate-level usage example to (hopefully) prevent folks from getting lost like I did.

[`Scrypt` struct's page]: https://docs.rs/scrypt/0.10.0/scrypt/struct.Scrypt.html


tarcieri:
Thanks!

Issue #327, opened 2022-07-31
sha512crypt index out of bounds


masmullin2000:
The function produce_byte_seq() will cause an index out of bounds panic if the password supplied to sha512_crypt has a length larger than 64 bytes.

kevoh1516:
I'm having the same issues when using pbkdf2 with sha256 and sha512.

tarcieri:
@kevoh1516 can you file a separate issue with a reproduction of your problem?

kevoh1516:
Sure.

tarcieri:
@masmullin2000 you didn't provide a reproduction for this issue.

There are test vectors for passwords which are 65, 127, 128, and 129-bytes long.

Closing this as lacking a reproduction and covered by the existing test vectors.

tarcieri:
Fixed in #328

PR #328, opened 2022-07-31
sha-crypt: enable passwords longer than 64 bytes in length

masmullin2000:
1) in produce_byte_seq, the fill_from slice will always be 64 bytes in
   length, due to the SHA512 algorithm digest supplying these bytes

2) produce_byte_seq() was attempting to index the fill_from slice beyond
   the 64 bytes when a password > 64 bytes in length was being used

signed-off-by: Michael Mullin <masmullin@gmail.com>

Addresses: https://github.com/RustCrypto/password-hashes/issues/327

### DIFF
--- a/sha-crypt/src/lib.rs
+++ b/sha-crypt/src/lib.rs
@@ -120,7 +120,7 @@ pub fn sha512_crypt(
 
     // 16.
     // Create byte sequence P.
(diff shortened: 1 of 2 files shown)


tarcieri:
Thank you!

PR #337, opened 2022-08-30
pbkdf2: add new wrapper functions

### LINKED ISSUE #335: add a convenient way for common pbkdf2::<Hmac<Sha1>>(*** )
eg:
```
let mut bytes = [0u8; 16];
pbkdf2::<Hmac<Sha1>>(password, b"saltysalt", 1003, bytes.as_mut());
```
go lang : https://pkg.go.dev/golang.org/x/crypto/pbkdf2
```
dk := pbkdf2.Key([]byte(password), []byte("saltysalt"), 1003, 16, sha1.New)
```
in rust, there are  additional info:  mut bytes , Hmac,
golang's pbkdf2.Key look like more convenient

newpavlov:
The ugly bounds on the HMAC functions can be improved by resolving https://github.com/RustCrypto/MACs/issues/104.

Closes #335.


pbkdf2/src/lib.rs
```diff
@@ -114,34 +144,136 @@ where
...
+/// instead of filling an input slice.
+///
+/// ```
+/// use hex_literal::hex;
+/// use pbkdf2::pbkdf2_array;
+/// use hmac::Hmac;
+/// use sha2::Sha256;
+///
+/// let res = pbkdf2_array::<Hmac<Sha256>, 20>(b"password", b"salt", 4096)
+///     .expect("HMAC can be initialized with any key length");
```
tarcieri, line 201: FWIW I've been trying to eliminate the usages of `expect` in documentation going forward, although the boilerplate to do so is a bit onerous:

```rust
#![cfg_attr(feature = "std", doc = "```")]
#![cfg_attr(not(feature = "std"), doc = "```ignore")]
//! # fn main() -> Result<(), Box<dyn std::error::Error>> {
//! [...]
//! # Ok(())
//! # }

tarcieri (approved):

tarcieri:
Looks like Cargo.lock needs updated but otherwise this seems really helpful!

pbkdf2/src/lib.rs
newpavlov: I think `expect` in this case is preferred to the error bubbling since the panic never will be triggered.

PR #347, opened 2022-10-11
wip: add sha256-crypt

ibotty:
This does not yet work correctly.  It fails some tests because the b64-functions don't work.  One byte is always wrong.  I triple-checked that I translated the table in https://akkadia.org/drepper/SHA-crypt.txt correctly (search for "base-64 encoded final C digest").  Unfortunately, I cannot find the error in the algorithm.

I would appreciate some help.

I will also have a look if it's possible to refactor the implementations so they share more code.

### DIFF
--- a/sha-crypt/src/b64.rs
+++ b/sha-crypt/src/b64.rs
@@ -1,14 +1,22 @@
 //! Base64 encoding support
 
-use crate::defs::{MAP, TAB};
+use crate::defs::{BLOCK_SIZE_SHA256, BLOCK_SIZE_SHA512, MAP_SHA256, MAP_SHA512, TAB};
 use alloc::vec::Vec;
 
 #[cfg(feature = "simple")]
 use crate::errors::DecodeError;
 
-pub fn encode(source: &[u8]) -> Vec<u8> {
+pub fn encode_sha512(source: &[u8]) -> Vec<u8> {
+    encode(source, MAP_SHA512)
+}
+
+pub fn encode_sha256(source: &[u8]) -> Vec<u8> {
+    encode(source, MAP_SHA256)
+}
+
+fn encode<const N: usize>(source: &[u8], map: [(u8, u8, u8, u8); N]) -> Vec<u8> {
     let mut out: Vec<u8> = vec![];
-    for entry in &MAP {
+    for entry in map {
         let mut w: usize = 0;
         if entry.3 > 2 {
             w |= source[entry.2 as usize] as usize;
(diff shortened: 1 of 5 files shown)


sha-crypt/tests/lib.rs
```diff
@@ -1,82 +1,100 @@
...
+        result_sha512:
             "svn8UoSVapNtMuq1ukKS4tPQd8iKwSMHWjl/O817G3uBnIFNjnQJuesI68u4OTLiBFdcbYEdFCoEOfaS35inz1",
         rounds: 5_000,
     },
     TestVector {
         input: "Hello world!",
         salt: "saltstringsaltstring",
-        result:
+        result_sha256: "3xv.VbSHBb41AL9AvLeujZkZRBAwqFMz2.opqey6IcA",
+        // result here: 3xv.VbSHBb41AL9AvLeujZkZRBAwqFMz2.opqey6Icw
```
ibotty, line 28: This is one of the faulty hashes.  Note the last character.

sha-crypt/src/b64.rs
```diff
@@ -67,8 +85,25 @@ mod tests {
...
+    #[test]
+    fn test_encode_decode_sha256() {
+        let original: [u8; 32] = [
+            0x0b, 0x5b, 0xdf, 0x7d, 0x92, 0xe2, 0xfc, 0xbd, 0xab, 0x57, 0xcb, 0xf3, 0xe0, 0x03,
+            0x16, 0x62, 0xd3, 0x6e, 0xa0, 0x57, 0x44, 0x8c, 0xca, 0x35, 0xec, 0x80, 0x75, 0x2a,
+            0x5f, 0xaf, 0x1a, 0xe5
+        ];
+
+        let e = super::encode_sha256(&original);
+        let d = super::decode_sha256(&e).unwrap();
```
ibotty, line 106: This test also fails.

Note, that when doing another roundtrip (encode +  decode), it converges instantly.

tarcieri:
@ibotty for what it's worth I don't completely understand the encoding used myself.

In the past I've tried to replace the bespoke implementation in `sha-crypt` with [`base64ct::Base64Crypt`](https://docs.rs/base64ct/latest/base64ct/struct.Base64Crypt.html).

But there's something I'm missing because a straightforward replacement of the current implementation fails.

I haven't read through the SHA-crypt.txt document in-depth enough to understand the exact peculiarities of the encoding.

ibotty:
afaict is [base64ct::Base64Crypt](https://docs.rs/base64ct/latest/base64ct/struct.Base64Crypt.html) missing the transpose maps.  It looks as if the trait does not support adding transpose maps.

Have a look at python's passlib API: https://foss.heptapod.net/python-libs/passlib/-/blob/branch/stable/passlib/utils/binary.py#L660
It is use by its sha2_crypts in https://foss.heptapod.net/python-libs/passlib/-/blob/branch/stable/passlib/handlers/sha2_crypt.py#L244

ibotty:
Checking [base64ct::Base64Crypt](https://docs.rs/base64ct/latest/base64ct/struct.Base64Crypt.html) I see, that it is not computing the same base64 hashes as python's passlib's `Hash64`.  Indeed, the test vectors in its testsuite are different from the ones that `passlib.utils.binary.h64.encode_bytes` generates.

So the missing transpose matrix is not the problem.

ibotty:
Ha, finally.  base64ct::Base64Crypt is big endian, while Hash64 is using little endian.  How to fix that, I don't know.

tarcieri:
Endianness of what? The inputs and outputs are bytestrings

ibotty:
That's what passlib is calling it.  I assume it's the 3-tuple of u8 that gets converted into the 4 base64 characters.
 https://passlib.readthedocs.io/en/stable/lib/passlib.utils.binary.html#base64engine-class and https://passlib.readthedocs.io/en/stable/lib/passlib.utils.binary.html#predefined-instances

ibotty:
I rebased on pr #349 which uses base64ct.  Now all tests pass.

WIP, because the other PRs are not yet merged and there is surely some code duplication that can be eliminated.

PR #349, opened 2022-10-13
WIP: use base64ct for sha-crypt base64 en-/decoding

ibotty:
This needs RustCrypto/formats#738 so it points the `base64ct` dependency to the PR's repo instead.  If accepted and a new release of `base64ct` is done, this should be updated.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6,7 +6,7 @@ version = 3
 name = "argon2"
 version = "0.5.0-pre"
 dependencies = [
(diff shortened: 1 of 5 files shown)


ibotty:
I did not check all unsuccessful checks, but most concern the PRs repository and look like flukes.

tarcieri:
@ibotty obsoleted by #350, correct?

ibotty:
Right!  Thank you for merging.

PR #350, opened 2022-10-13
wp: add sha256_crypt

ibotty:
This uses #349 and and is missing some code dedup.

### DIFF
--- a/sha-crypt/Cargo.toml
+++ b/sha-crypt/Cargo.toml
@@ -21,6 +21,7 @@ sha2 = { version = "0.10", default-features = false }
 # optional dependencies
 rand = { version = "0.8", optional = true }
 subtle = { version = ">=2, <2.5", optional = true, default-features = false }
+base64ct = "1.5.3"
 
 [features]
 default = ["simple"]
--- a/sha-crypt/src/b64.rs
+++ b/sha-crypt/src/b64.rs
@@ -1,64 +1,41 @@
 //! Base64 encoding support
 
-use crate::defs::{MAP, TAB};
+use crate::defs::{BLOCK_SIZE, MAP_SHA512, PW_SIZE_SHA512};
 use alloc::vec::Vec;
+use base64ct::{Base64ShaCrypt, Encoding};
+
 
 #[cfg(feature = "simple")]
 use crate::errors::DecodeError;
 
-pub fn encode(source: &[u8]) -> Vec<u8> {
-    let mut out: Vec<u8> = vec![];
-    for entry in &MAP {
-        let mut w: usize = 0;
-        if entry.3 > 2 {
-            w |= source[entry.2 as usize] as usize;
-            w <<= 8;
-            w |= source[entry.1 as usize] as usize;
-            w <<= 8;
-        }
-        w |= source[entry.0 as usize] as usize;
-
-        for _ in 0..entry.3 {
-            out.push(TAB[(w & 0x3f) as usize]);
-            w >>= 6;
-        }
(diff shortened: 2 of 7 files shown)


ibotty (commented):
I'd prefer the tables to be in the format as in the spec.  What do you think?

sha-crypt/Cargo.toml
```diff
@@ -21,6 +21,7 @@ sha2 = { version = "0.10", default-features = false }
 # optional dependencies
 rand = { version = "0.8", optional = true }
 subtle = { version = ">=2, <2.5", optional = true, default-features = false }
+base64ct = { git = "https://github.com/ibotty/rustcrypto-formats.git", branch = "shacrypt" }
```
newpavlov, line 24: You do not need `.git` here.

newpavlov (commented):
>I'd prefer the tables to be in the format as in the spec

I guess it would make checking the constant easier. You can use `#[rustfmt::skip]`.

sha-crypt/Cargo.toml
ibotty: This will be gone if and when a new base64ct version is released.  Or do you think this is actively harmful?

newpavlov: I meant you should use `https://github.com/ibotty/rustcrypto-formats`, without the `.git` part.

ibotty: Oh, it worked locally so I assumed that was not a problem.  I will change and rebase in a sec.

newpavlov: Also, it looks like you created the formats PR against the `use-base64ct` branch, not `shacrypt`.

newpavlov: Hm, removing `.git` has not solved the CI issue. I am not sure why it does not work... Maybe try try to use commit hash instead of branch name?

sha-crypt/src/b64.rs
```diff
@@ -1,64 +1,64 @@
...
-    for entry in &MAP {
-        let mut w: usize = 0;
-        if entry.3 > 2 {
-            w |= source[entry.2 as usize] as usize;
-            w <<= 8;
-            w |= source[entry.1 as usize] as usize;
-            w <<= 8;
-        }
-        w |= source[entry.0 as usize] as usize;
+pub fn encode_sha512(source: &[u8]) -> Vec<u8> {
```
tarcieri, line 12: FWIW it'd be really great to get rid of `Vec` here if possible. This is one of the few crates we have which has a hard dependency on `alloc`.

ibotty: I don't have experience in no-std libraries.  I'll see what I can do, but I won't spend too much effort on this.

tarcieri: Here it should be fairly easy to get rid of by just returning the array, rather than converting it into a `Vec`

ibotty: The other vectors look harder to eliminate though.

tarcieri: No worries then!

ibotty:
Can you ping me when you released a new base64ct?  I'll then update the cargo file and we can move forward with it.

tarcieri:
@ibotty `base64ct` v1.5.3 released

ibotty:
Thanks!  I rebased and cleaned up some clippy warnings.  The tests look alright.  What do you think?

tarcieri (approved):
Thanks! This looks good now.

ibotty:
Thank you!

PR #363, opened 2022-12-19
pbkdf2: use `sha1` dependency directly

koushiro:


koushiro:
@tarcieri PTAL

tarcieri:
Thanks!

PR #372, opened 2023-01-29
`bcrypt-pbkdf` without `alloc`

antonok-edm:
I'm using `bcrypt-pbkdf` in an embedded context without a global allocator. To support this, I added a new function `bcrypt_pbkdf_with_memory`, which takes in an additional memory buffer argument (similar to some of the other crates in this repo with an optional alloc feature).


tarcieri (approved):
This seems simple enough. Thanks!

PR #374, opened 2023-02-04
Disable default feature (i.e. alloc) of zeroize in argon2

rlee287:
zeroize is not used on Vec or String types for argon2, so removing the default alloc feature allows the zeroize feature to be enabled in argon2 for no-std targets.


tarcieri:
Thanks! Seems like an oversight

PR #375, opened 2023-02-15
pbkdf2: make public PBKDF2 Ident values

ChanTsune:
For RustCrypto members

Thank you for developing a awesome crate!

I have a question.

Why do the other crates maintained by this repository have published Ident values, but not in `pbkdf2`?

Looking at the code, it seems that there is an intention to publish it, but it is not available because there is no `use` declaration.

If you just forgot to publish it, I would appreciate it if you could merge this PR and make it available.

### DIFF
--- a/pbkdf2/src/lib.rs
+++ b/pbkdf2/src/lib.rs
@@ -95,10 +95,7 @@ pub use password_hash;
 mod simple;
 
 #[cfg(feature = "simple")]
-pub use crate::simple::{Algorithm, Params, Pbkdf2, PBKDF2_SHA256_IDENT, PBKDF2_SHA512_IDENT};
-
-#[cfg(all(feature = "simple", feature = "sha1"))]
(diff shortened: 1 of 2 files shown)


ChanTsune:
Thanks for your advice.
I added _IDENT as a suffix to each.

tarcieri:
Personally I think I'd prefer the inherent constant approach, since it cuts down on the number of exports/imports

ChanTsune:
OK!, so how about making it an intrinsic constant of `Algorithm` following the other crates as follows?

https://github.com/ChanTsune/password-hashes/blob/ca2ff2abdfa347ce7d6a3e4bd764f988ba9a282c/pbkdf2/src/simple.rs#L82-L94

tarcieri:
That looks good, yes

tarcieri (approved):
Looks good, thanks!

PR #379, opened 2023-03-04
argon2: provide std::error::Error::source() for Error

z33ky:


argon2/src/error.rs
```diff
@@ -121,4 +121,25 @@ impl From<Error> for password_hash::Error {
...
+            Self::OutputTooLong => None,
+            Self::PwdTooLong => None,
+            Self::SaltTooShort => None,
+            Self::SaltTooLong => None,
+            Self::SecretTooLong => None,
+            Self::ThreadsTooFew => None,
+            Self::ThreadsTooMany => None,
+            Self::TimeTooSmall => None,
+            Self::VersionInvalid => None,
+        }
```
tarcieri, line 143: Really this can just be...

```suggestion
        match self {
            Self::B64Encoding(err) => Some(err),
            _ => None
        }
```

...and now that I mention it, this is also true of https://github.com/RustCrypto/traits/pull/1264/files#diff-c39a2e2913ce6aa08f2b5ae5c168f57989e954720d0bc3e587a893ce07dc2e3a

z33ky: Indeed, I can simplify it.
A potential downside is that added variants with wrapped errors will by default return `None`, whereas with this manual listing new variants will trigger an error until added on purpose.

tarcieri: There is that, I suppose

tarcieri:
@z33ky can you `commit --amend` and force push to retrigger the build?

PR #416, opened 2023-05-20
Annotate `Algorithm::as_str` returning static str

FirelightFlagboy:
Previously `Algorithm::as_str` used no lifetime specifier on the returned `as_str`, so the returned value inherited the lifetime of `&self`.

But the returned value are static value and could use the lifetime `'static` instead.


tarcieri (approved):

tarcieri:
Thanks!

PR #423, opened 2023-06-13
Argon2 impl Debug

RoDmitry:


tarcieri:
The `Argon2` struct contains a `secret` field which a blanket `Debug` impl like this will reveal.

Instead it needs a handwritten one which uses `debug_struct` and omits the `secret` field, either substituting e.g. an ellipsis `"..."` for the `secret` field or using `finish_non_exhaustive`.

RoDmitry:
Done using `finish_non_exhaustive`

PR #435, opened 2023-07-01
scrypt: Update docs for recommended log_n parameter

Granddave:
When updating the parameters in a previous PR (#388) this documentation was missed.

### DIFF
--- a/scrypt/src/params.rs
+++ b/scrypt/src/params.rs
@@ -87,7 +87,7 @@ impl Params {
         })
     }
 
-    /// Recommended values sufficient for most use-cases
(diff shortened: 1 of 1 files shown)


scrypt/src/params.rs
```diff
@@ -88,7 +88,7 @@ impl Params {
     }
 
     /// Recommended values sufficient for most use-cases
```
tarcieri, line 90: Perhaps add a link to the OWASP cheat sheet here:

```suggestion
    /// Recommended values according to the [OWASP cheat sheet](https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html#scrypt)
```

tarcieri:
Thanks!

PR #437, opened 2023-07-03
argon2: add benchmarks using criterion

dyc3:
I added some benchmarks using the `criterion` crate. The reason I chose to include criterion is because it makes it easier to determine if a change has a statistically significant impact.

I'm not really sure which benchmarks will produce good signal for us, so the parameters will probably need to be tweaked before merging. I'm open to suggestions.


newpavlov:
I would prefer to keep criterion benchmarks in a separate crate as we do in the [stream ciphers](https://github.com/RustCrypto/stream-ciphers) repository to prevent criterion being pulled in our CI jobs.

Cargo.toml
```diff
@@ -4,6 +4,7 @@ members = [
     "argon2",
     "balloon-hash",
     "bcrypt-pbkdf",
+    "benches",
```
newpavlov, line 7: No, keep the `benches` crate outside of the common workspace. Otherwise its dependencies become part of the workspace which we do not want. You may need to add `**/Cargo.lock` to `.gitignore` to prevent `bencehs/Cargo.lock` from being committed.

benches/Cargo.toml
```diff
@@ -0,0 +1,18 @@
...
+[dependencies]
+
+[dev-dependencies]
+argon2 = { path = "../argon2" }
+criterion = { version = "0.4", features = ["html_reports"] }
+
+[[bench]]
+name = "argon2"
+path = "src/argon2.rs"
+harness = false
```
newpavlov, line 18: Add a new line here.

newpavlov (approved):
Thank you!

PR #438, opened 2023-07-03
argon2: Add const compatibility

### LINKED ISSUE #436: argon2: const ParamsBuilder
Can the ParamsBuilder methods be made const?
Const fn's would help out, because then I could make the Params object a Compile time const.
I took a little look at the code and saw no obvious reason why it should not be possible.

```rust
const PARAMS:argon2::Params = match argon2::ParamsBuilder::new().build() {
    Ok(v) => v,
    Err(_) => panic!("Failed to create argon2 params"),
};
```

https://github.com/RustCrypto/password-hashes/blob/master/argon2/src/params.rs#L367

C0D3-M4513R:
Closes #436

### DIFF
--- a/argon2/Cargo.toml
+++ b/argon2/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "argon2"
-version = "0.5.1"
+version = "0.5.0"
 description = """
 Pure Rust implementation of the Argon2 password hashing function with support
 for the Argon2d, Argon2i, and Argon2id algorithmic variants
--- a/argon2/Cargo.toml
+++ b/argon2/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "argon2"
-version = "0.5.0"
+version = "0.5.1"
 description = """
 Pure Rust implementation of the Argon2 password hashing function with support
 for the Argon2d, Argon2i, and Argon2id algorithmic variants
--- a/argon2/src/params.rs
+++ b/argon2/src/params.rs
@@ -198,10 +198,6 @@ impl Params {
     pub fn block_count(&self) -> usize {
         self.segment_length() * self.lanes() * SYNC_POINTS
     }
-    /// Just like [Default::default], but const.
-    pub const fn default_const() -> Params {
-        Params::DEFAULT
-    }
 }
 
 impl Default for Params {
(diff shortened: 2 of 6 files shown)


argon2/src/params.rs
```diff
@@ -224,6 +227,14 @@ macro_rules! param_buf {
                 Ok(Self { bytes, len })
             }
 
+            #[doc = "Just like [Default::default], but const."]
```
newpavlov, line 230: Why do you use the `#[doc = "..."]` attribute here instead of the usual `///`?

argon2/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "argon2"
-version = "0.5.0"
+version = "0.5.1"
```
tarcieri, line 3: Please don't bump the version

argon2/src/params.rs
```diff
@@ -198,6 +198,10 @@ impl Params {
     pub fn block_count(&self) -> usize {
         self.segment_length() * self.lanes() * SYNC_POINTS
     }
+    /// Just like [Default::default], but const.
+    pub const fn default_const() -> Params {
+        Params::DEFAULT
+    }
```
tarcieri, line 204: This is unnecessary now. You can just use `Params::DEFAULT`.

argon2/src/params.rs
```diff
@@ -233,6 +236,14 @@ macro_rules! param_buf {
                 Ok(Self { bytes, len })
             }
 
+            /// Just like [Default::default], but const.
+            pub const fn default_const() -> Self {
+                Self {
+                    bytes: [0u8; Self::MAX_LEN],
+                    len: 0,
+                }
+            }
```
tarcieri, line 241: This could instead be:

```suggestion
            /// Empty value.
            pub const EMPTY: Self = Self {
                bytes: [0u8; Self::MAX_LEN],
                len: 0,
            };
```

tarcieri: Also this could be used in the definition of `Params::DEFAULT`

tarcieri (approved):

tarcieri:
It seems I merged this after #440 which is not compatible with these changes, namely it introduced runtime CPU feature detection which is not const-friendly.

I'm going to revert for now.

PR #439, opened 2023-07-03
argon2: add `Params::DEFAULT` constant

tarcieri:
Allows for accessing default parameters in const contexts


tarcieri:
That seems at odds with the notion of a default, especially in this context: we have specially selected the OWASP recommended parameters specifically so you *don't* have to customize them.

If you want customization, by definition you are not using the default.

Issue #104, opened 2021-01-29
argon2: optimized implementation


tarcieri:
The current `argon2` crate implementation is a translation of `ref.c` from the reference implementation:

https://github.com/P-H-C/phc-winner-argon2/blob/92cd2e1/src/ref.c

It could be improved by translating `opt.c` instead, which provides e.g. SIMD support:

https://github.com/P-H-C/phc-winner-argon2/blob/92cd2e1/src/opt.c

cooperwalbrun:
I came here to comment that I think this would be a great improvement to the library. I have been experimenting with the latest version of the `argon2` crate recently and I noticed some slow execution times using `hash_password`. However, I am not an expert on this crate (or even cryptography in general), so I am not sure if this is expected?

In one of my unit tests, I perform `hash_password("text".as_bytes(), ...)` one time with the following parameters:

* `ｍ=64`
* `t=3`
* `p=1`
* `l=32`

on an i7-6700K CPU @ 4.00GHz. I consistently observe about 4.15-4.25 second execution times of this test. When I lower the parameters to `m=32` and `t=2`, I still see 1.35-1.40 second execution times, and with even lower values I can make it faster still, as I would expect. But these times strike me as being on the high side relative to their parameters. Even with `t=1` and `m=1`, I observe times significantly slower than [the benchmarks using the same values reported by the original C code](https://github.com/P-H-C/phc-winner-argon2#benchmarks).

I think it is also worth mentioning that [the specification itself](https://www.cryptolux.org/images/0/0d/Argon2.pdf) recommends using parameter values that lead to the hashing taking about 0.5 seconds on average for my use case (page 16: backend server authentication) - for me, this involves specifying values that are actually *below* the [best-practice guidelines](https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html), thus demonstrating that I cannot use `argon2`, or at least this crate's implementation of it. If that is my reality, it is what it is I suppose.

I did some internet searching and I saw that others have previously questioned the efficiency of Argon2's implementation in Rust in general (not just this crate), so is it possible that this is merely a consequence of not using a SIMD-based implementation? See:

[[1]](https://www.reddit.com/r/rust/comments/xjw6h5/comment/ipcm030/?utm_source=share&utm_medium=web2x&context=3)
>I tested both crates and found that while their performance is very similar (with `argon2` ~5% faster), in comparison to reference C implementation it's around 1.7x slower for some reason.

[...]

PR #440, opened 2023-07-05
argon2: optimize with AVX2 SIMD

dyc3:
I've tried my hardest to get this to work, but I think I'm in a little over my head. The most I've managed to do is port the [macros from the optimized reference version](https://github.com/P-H-C/phc-winner-argon2/blob/f57e61e19229e23c4445b85494dbf7c07de721cb/src/blake2/blamka-round-opt.h).

~~My main problem is that I have no idea how these indexes into `state` work:~~

~~https://github.com/P-H-C/phc-winner-argon2/blob/92cd2e1/src/opt.c#L94-L102~~

If someone could give me a hand, I'd be happy to fix/finish this. This is the first time I've ever tried to do stuff with SIMD, so bear with me as I'm still learning.

related to #104

### DIFF
--- a/argon2/Cargo.toml
+++ b/argon2/Cargo.toml
@@ -18,11 +18,13 @@ rust-version = "1.65"
 [dependencies]
 base64ct = "1"
 blake2 = { version = "0.10.6", default-features = false }
-cpufeatures = "0.2.9"
 
 # optional dependencies
 password-hash = { version = "0.5", optional = true }
 zeroize = { version = "1", default-features = false, optional = true }
+
+[target.'cfg(any(target_arch = "x86", target_arch = "x86_64", target_arch = "aarch64"))'.dependencies]
+cpufeatures = "0.2.9"
 
 [dev-dependencies]
 hex-literal = "0.4"
--- a/argon2/Cargo.toml
+++ b/argon2/Cargo.toml
@@ -23,7 +23,7 @@ blake2 = { version = "0.10.6", default-features = false }
 password-hash = { version = "0.5", optional = true }
 zeroize = { version = "1", default-features = false, optional = true }
 
-[target.'cfg(any(target_arch = "x86", target_arch = "x86_64", target_arch = "aarch64"))'.dependencies]
+[target.'cfg(any(target_arch = "x86", target_arch = "x86_64"))'.dependencies]
 cpufeatures = "0.2.9"
 
 [dev-dependencies]
--- a/argon2/Cargo.toml
+++ b/argon2/Cargo.toml
@@ -18,6 +18,7 @@ rust-version = "1.65"
 [dependencies]
 base64ct = "1"
 blake2 = { version = "0.10.6", default-features = false }
+cpufeatures = "0.2.9"
 
 # optional dependencies
 password-hash = { version = "0.5", optional = true }
--- a/argon2/src/block.rs
+++ b/argon2/src/block.rs
@@ -44,172 +44,6 @@ macro_rules! permute {
     };
 }
 
-const fn _mm_shuffle2(z: i32, y: i32, x: i32, w: i32) -> i32 {
-    (z << 6) | (y << 4) | (x << 2) | w
-}
-
-macro_rules! rotr32 {
-    ($x:expr) => {
-        _mm256_shuffle_epi32($x, _mm_shuffle2(2, 3, 0, 1))
-    };
-}
-
-macro_rules! rotr24 {
-    ($x:expr) => {
-        _mm256_shuffle_epi8(
-            $x,
-            _mm256_setr_epi8(
-                3, 4, 5, 6, 7, 0, 1, 2, 11, 12, 13, 14, 15, 8, 9, 10, 3, 4, 5, 6, 7, 0, 1, 2, 11,
-                12, 13, 14, 15, 8, 9, 10,
-            ),
-        )
-    };
-}
-
-macro_rules! rotr16 {
-    ($x:expr) => {
-        _mm256_shuffle_epi8(
-            $x,
-            _mm256_setr_epi8(
-                2, 3, 4, 5, 6, 7, 0, 1, 10, 11, 12, 13, 14, 15, 8, 9, 2, 3, 4, 5, 6, 7, 0, 1, 10,
-                11, 12, 13, 14, 15, 8, 9,
-            ),
-        )
-    };
-}
-
-macro_rules! rotr63 {
-    ($x:expr) => {
-        _mm256_xor_si256(_mm256_srli_epi64($x, 63), _mm256_add_epi64($x, $x))
-    };
-}
-
-macro_rules! G1_AVX2 {
-    ($A0:expr, $A1:expr, $B0:expr, $B1:expr, $C0:expr, $C1:expr, $D0:expr, $D1:expr) => {{
-        let ml = _mm256_mul_epu32($A0, $B0);
-        let ml = _mm256_add_epi64(ml, ml);
-        $A0 = _mm256_add_epi64($A0, _mm256_add_epi64($B0, ml));
-        $D0 = _mm256_xor_si256($D0, $A0);
-        $D0 = rotr32!($D0);
-        let ml = _mm256_mul_epu32($C0, $D0);
-        let ml = _mm256_add_epi64(ml, ml);
-        $C0 = _mm256_add_epi64($C0, _mm256_add_epi64($D0, ml));
-        $B0 = _mm256_xor_si256($B0, $C0);
-        $B0 = rotr24!($B0);
-        let ml = _mm256_mul_epu32($A1, $B1);
-        let ml = _mm256_add_epi64(ml, ml);
-        $A1 = _mm256_add_epi64($A1, _mm256_add_epi64($B1, ml));
-        $D1 = _mm256_xor_si256($D1, $A1);
-        $D1 = rotr32!($D1);
-        let ml = _mm256_mul_epu32($C1, $D1);
-        let ml = _mm256_add_epi64(ml, ml);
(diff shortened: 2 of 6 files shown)


dyc3:
Ok, I was able to verify that I ported all the macros correctly, and I fixed some stuff. I'm still not entirely sure where I'm going wrong here, as the tests are still failing.

dyc3:
I think it may be possible that I've encountered a compiler bug. When calling the intrinsic `_mm256_loadu_si256`, it seems like rustc does not keep track of which registers it has used.

See `example::Block::compress` in here: https://godbolt.org/z/1Gz3aM8xa

dyc3:
I found out there was a way to tell the compiler to generate code based on a target feature being enabled with the `#[target_feature(enable = "avx2")]` attribute. With this method, it likely be safe to do the same thing with avx512, but I don't have a CPU that supports it to make sure.

https://godbolt.org/z/843z7eGYb demonstrates the code gen using this method, which is implemented in [70e34de](https://github.com/RustCrypto/password-hashes/pull/440/commits/70e34def273b35a6d95f3e12c26cf9e700b54d57). 

Running benchmarks on my Ryzen 9 3900X, it shows about a 20-30% performance improvement. Would this solution be acceptable? I can clean up the commit history here if it matters.

argon2/src/block.rs
```diff
@@ -67,6 +71,55 @@ impl Block {
     }
 
     pub(crate) fn compress(rhs: &Self, lhs: &Self) -> Self {
+        #[cfg(any(target_arch = "x86_64"))]
+        {
+            let (_, avx2) = avx2_cpuid::init_get();
+            if avx2 {
```
tarcieri, line 77: This is calling `CPUID` for every invocation of `compress`. It'd be a lot better to find somewhere to store the init token (possibly in the `Argon2` struct, moving the `compress` function there, or at least the decision to use AVX2 or not).

Failing that you don't need to fetch and discard it:

```suggestion
            if avx2_cpuid::get() {
```

argon2/src/block.rs
```diff
@@ -67,6 +71,55 @@ impl Block {
...
+                return unsafe { Self::compress_avx2(rhs, lhs) };
+            }
+        }
+        Self::compress_safe(rhs, lhs)
+    }
+
+    fn compress_safe(rhs: &Self, lhs: &Self) -> Self {
```
tarcieri, line 84: We more typically use `soft` or `portable` to refer to these kinds of intrinsic-free implementations, rather than `safe`

argon2/src/block.rs
newpavlov: `init_get` does not call CPUID every time, it properly caches results from the first call. ZST token acts as an initialization prove, which allows us to remove the third branch used for inner static initialization.

argon2/Cargo.toml
```diff
@@ -23,6 +23,9 @@ blake2 = { version = "0.10.6", default-features = false }
 password-hash = { version = "0.5", optional = true }
 zeroize = { version = "1", default-features = false, optional = true }
 
+[target.'cfg(any(target_arch = "x86", target_arch = "x86_64", target_arch = "aarch64"))'.dependencies]
```
newpavlov, line 26: You don't use `cpufeatures` on AArch64, so you can remove it from this `cfg`.

argon2/src/block.rs
```diff
@@ -44,8 +44,12 @@ macro_rules! permute {
     };
 }
 
+#[cfg(any(target_arch = "x86_64"))]
+cpufeatures::new!(avx2_cpuid, "avx2");
+
 /// Structure for the (1 KiB) memory block implemented as 128 64-bit words.
 #[derive(Copy, Clone, Debug)]
+#[cfg_attr(test, derive(PartialEq))]
```
newpavlov, line 52: I think it will be better to write `assert_eq!(result.0, result_av2.0);` than use this `cfg`.

argon2/src/block.rs
```diff
@@ -67,6 +71,55 @@ impl Block {
     }
 
     pub(crate) fn compress(rhs: &Self, lhs: &Self) -> Self {
+        #[cfg(any(target_arch = "x86_64"))]
```
newpavlov, line 74: You forgot `x86` here.

argon2/src/block.rs
```diff
@@ -144,3 +197,32 @@ impl Zeroize for Block {
...
+    fn compress_avx2() {
+        let lhs = Block([
+            0, 0, 0, 2048, 4, 2, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+            0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+            0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+            0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+            0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+        ]);
+        let rhs = Block([
+            0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
```
newpavlov, line 216: Use `[0; N]` to initialize this block.

argon2/src/block.rs
```diff
@@ -144,3 +197,32 @@ impl Zeroize for Block {
...
+            0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+            0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+            0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+            0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+        ]);
+
+        let result = Block::compress_safe(&rhs, &lhs);
+        let result_av2 = unsafe { Block::compress_avx2(&rhs, &lhs) };
```
newpavlov, line 224: I think you meant `result_avx2`.

argon2/src/block.rs
```diff
@@ -144,3 +197,32 @@ impl Zeroize for Block {
...
+
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    #[cfg(target_arch = "x86_64")]
+    #[test]
+    fn compress_avx2() {
+        let lhs = Block([
+            0, 0, 0, 2048, 4, 2, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
```
newpavlov, line 209: It would be more concise to initialize block with zeros and then `copy_from_slice` the first 7 values.

newpavlov:
So the compress function implementations are identical and the only difference is `#[target_feature(enable = "avx2")]`, correct? I don't think we need to duplicate the function implementations. You could try to mark the compress function as `#[inline(always)]` and then introduce an AVX2 wrapper around it.

argon2/src/block.rs
newpavlov: They are not needed. `cpufeatures` already efficiently handles it.

dyc3: Ah, I see. So does that mean this is fine as is?

dyc3: So does that mean that this is fine as is? Or do we still need to keep the init token around.

tarcieri: > ZST token acts as an initialization proof

...but it's being thrown away here

argon2/src/block.rs
```diff
@@ -101,6 +106,12 @@ impl Block {
         q ^= &r;
         q
     }
+
+    #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+    #[target_feature(enable = "avx2")]
+    pub(crate) unsafe fn compress_avx2(rhs: &Self, lhs: &Self) -> Self {
+        Self::compress_soft(rhs, lhs)
+    }
```
tarcieri, line 114: It feels like this is all an implementation detail that could be factored somewhere else, e.g. nested inner functions

dyc3: Did you have something specific in mind? I'm not quite seeing what you're saying.

argon2/src/block.rs
newpavlov: >but it's being thrown away here

Yes, but lack of initialization proof does not mean that we disable caching of CPUID results. `avx2_cpuid::(init_)get()` simply has to [account](https://github.com/RustCrypto/utils/blob/master/cpufeatures/src/lib.rs#L149-L166) for possibility that the inner static could be uninitialized. While `token.get()` can [assume](https://github.com/RustCrypto/utils/blob/master/cpufeatures/src/lib.rs#L137-L143) that the static was already initialized (thus instead of 3 branches, the code will have only 2).

argon2/src/block.rs
tarcieri: I can take care of it in a followup

PR #444, opened 2023-07-12
argon2: fold `compress_avx2` into an inner function

tarcieri:
This changes `Argon2::compress` to contain all AVX2-related logic, so it doesn't bleed into the `Block` type yet (especially since that lacks any AVX2-specific implementation).

### DIFF
--- a/argon2/src/block.rs
+++ b/argon2/src/block.rs
@@ -66,8 +66,10 @@ impl Block {
         unsafe { &mut *(self.0.as_mut_ptr() as *mut [u8; Self::SIZE]) }
     }
 
+    /// NOTE: do not call this directly. It should only be called via
+    /// `Argon2::compress`.
     #[inline(always)]
-    pub(crate) fn compress_soft(rhs: &Self, lhs: &Self) -> Self {
+    pub(crate) fn compress(rhs: &Self, lhs: &Self) -> Self {
         let r = *rhs ^ lhs;
 
         // Apply permutations rowwise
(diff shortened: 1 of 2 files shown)


tarcieri:
Followup to #440

cc @dyc3 @newpavlov

argon2/src/block.rs
```diff
@@ -66,8 +66,10 @@ impl Block {
         unsafe { &mut *(self.0.as_mut_ptr() as *mut [u8; Self::SIZE]) }
     }
 
+    /// NOTE: do not call this directly. It should only be called via
+    /// `Argon2::compress`.
     #[inline(always)]
```
tarcieri, line 71: I suspect this would be fine as `#[inline]` but haven't benchmarked yet

newpavlov (approved):

argon2/src/block.rs
newpavlov: AFAIK `#[inline]` is effectively useless for crate-private functions. You could remove the attribute completely  to check whether compiler decides to inline it by itself. But to be on the safe side and to be more robust in the face of future compiler changes, I think it's fine to use `#[inline(always)]`.

argon2/src/block.rs
```diff
@@ -151,21 +147,3 @@ impl Zeroize for Block {
...
-        let mut lhs = Block([0; 128]);
-        lhs.0[0..7].copy_from_slice(&[0, 0, 0, 2048, 4, 2, 1]);
-        let rhs = Block([0; 128]);
-
-        let result = Block::compress_soft(&rhs, &lhs);
-        let result_avx2 = unsafe { Block::compress_avx2(&rhs, &lhs) };
-
-        assert_eq!(result.0, result_avx2.0);
-    }
-}
```
tarcieri, line 171: This test feels a bit superfluous as for now the implementation is just `Block::compress` with optimizations enabled

dyc3 (approved):
Ah, now I see what you were saying before. This looks good to me.

PR #450, opened 2023-08-11
argon2: more const-ness

C0D3-M4513R:
Redo of #438 after #440.
Rebased original PR, then applied fixes and then sqashed

### DIFF
--- a/argon2/src/params.rs
+++ b/argon2/src/params.rs
@@ -110,7 +110,14 @@ impl Params {
         p_cost: u32,
         output_len: Option<usize>,
     ) -> Result<Self> {
-        ParamsBuilder::new_params(m_cost, t_cost, p_cost, None, None, output_len).build()
+        let mut builder = ParamsBuilder::new()
+            .m_cost_const(m_cost)
+            .t_cost_const(t_cost)
+            .p_cost_const(p_cost);
+        if let Some(output_len) = output_len {
+            builder = builder.output_len_const(output_len);
+        }
+        builder.build()
     }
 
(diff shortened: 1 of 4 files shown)


argon2/src/params.rs
```diff
@@ -384,8 +387,27 @@ pub struct ParamsBuilder {
...
+    ) -> Self {
+        Self {
+            m_cost,
+            t_cost,
+            p_cost,
+            keyid,
+            data,
+            output_len,
+        }
     }
```
tarcieri, line 411: This seems like it defeats the point of having a builder. The reason it exists is so you don't have to pass a big bag of parameters like this.

tarcieri: But then why have a builder?

tarcieri:
I think this PR could've benefitted from some up-front design discussion.

I'm not sure it necessarily makes sense to constify the builder. And doing so, at least the way you propose, means copying the entire thing every time you use any builder method. There are some tradeoffs there I think are worth discussing.

tarcieri:
> Well I don't want to duplicate any code from the Builder's build method in the Param's new method.

Okay, so it seems like part of the problem for now is `Params::new` is using the builder internally in its current implementation.

But `Params::new` is much more ideal for constification since it already accepts all of the parameters explicitly.

Perhaps instead of trying to constify the `Builder`, you could eliminate the internal use of it within `Params` and make `const fn new` which could then in turn be used by the `Builder`, rather than the other way around.

tarcieri (approved):

tarcieri:
Thank you!

PR #452, opened 2023-08-12
Argon2: Fixed max params

### LINKED ISSUE #451: Argon2: Max params
* Max m_cost is (2^32)-1 "4 TiB" but it's defined as (2^28)-1 "256 GiB":
https://github.com/RustCrypto/password-hashes/blob/7ff9080a4ead680037acd43928f4815d0bb13f7c/argon2/src/params.rs#L15-L18 vs https://github.com/RustCrypto/password-hashes/blob/7ff9080a4ead680037acd43928f4815d0bb13f7c/argon2/src/params.rs#L49

* Max p_cost is (2^24)-1 but it's stated as 255:
https://github.com/RustCrypto/password-hashes/blob/7ff9080a4ead680037acd43928f4815d0bb13f7c/argon2/src/params.rs#L25-L28 vs https://github.com/RustCrypto/password-hashes/blob/7ff9080a4ead680037acd43928f4815d0bb13f7c/argon2/src/params.rs#L67

Sc00bz:
Fixes #451

### DIFF
--- a/argon2/src/params.rs
+++ b/argon2/src/params.rs
@@ -22,9 +22,9 @@ pub struct Params {
     /// Value is an integer in decimal (1 to 10 digits).
     t_cost: u32,
 
-    /// Degree of parallelism, between 1 and 255.
+    /// Degree of parallelism, between 1 and (2^24)-1.
     ///
-    /// Value is an integer in decimal (1 to 3 digits).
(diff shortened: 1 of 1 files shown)


Sc00bz:
Sorry for the multiple commits. I was like this is a simple change "what could go wrong?"

tarcieri:
Thanks!

PR #458, opened 2023-09-05
Argon2: Fix `Params` docs

sorairolake:
* The minimum value of `m_cost` is 8 ( 8*`MIN_OUTPUT_LEN`), not 1.
* The maximum value of `p_cost` is (2^24)-1, not 255.

These are defined in [RFC 9106](https://datatracker.ietf.org/doc/html/rfc9106), and [`MIN_M_COST`](https://docs.rs/argon2/0.5.2/argon2/struct.Params.html#associatedconstant.MIN_M_COST) and [`MAX_P_COST`](https://docs.rs/argon2/0.5.2/argon2/struct.Params.html#associatedconstant.MAX_P_COST) are also like that.

See also: #451, #452

### DIFF
--- a/argon2/src/params.rs
+++ b/argon2/src/params.rs
@@ -12,7 +12,7 @@ use password_hash::{ParamsString, PasswordHash};
 /// These are parameters which can be encoded into a PHC hash string.
 #[derive(Clone, Debug, Eq, PartialEq)]
 pub struct Params {
-    /// Memory size, expressed in kibibytes, between 1 and (2^32)-1.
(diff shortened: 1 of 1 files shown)


tarcieri (approved):
I verified these match the RFC.

sorairolake:
`Params::new` verifies that `m_cost` is greater than or equal to `p_cost`*8.

https://github.com/RustCrypto/password-hashes/blob/144b2e41a686524fb8f971454cf65a7fbcdf7bed/argon2/src/params.rs#L119-L121

So, it might be better to explain that the minimum value of `m_cost` is `p_cost`*8 instead of 8.

Issue #380, opened 2023-03-04
argon2: parallel memory view abstraction


tarcieri:
The previous parallel implementation of Argon2 was removed in #247. There were soundness concerns about how it handled mutable aliasing, namely that it's unsound for mutable aliases to exist even if they aren't accessed.

This is a discussion issue for how to build a sound abstraction for use in a new parallel implementation of Argon2. Please make sure to read through this issue before attempting a new parallel implementation.

## Argon2 memory layout

The memory layout used by Argon2 is somewhat difficult to model in Rust due to how it's arranged, though this choice of arrangement is due to its goals of sequential memory hardness.

https://datatracker.ietf.org/doc/html/rfc9106#section-3.4

```
3.4. Indexing

To enable parallel block computation, we further partition the memory
matrix into SL = 4 vertical slices. The intersection of a slice and a
lane is called a segment, which has a length of q/SL. Segments of the
same slice can be computed in parallel and do not reference blocks
from each other. All other blocks can be referenced.

    slice 0    slice 1    slice 2    slice 3
    ___/\___   ___/\___   ___/\___   ___/\___
   /        \ /        \ /        \ /        \
  +----------+----------+----------+----------+
  |          |          |          |          | > lane 0
  +----------+----------+----------+----------+
  |          |          |          |          | > lane 1
  +----------+----------+----------+----------+
  |          |          |          |          | > lane 2
  +----------+----------+----------+----------+
  |         ...        ...        ...         | ...
  +----------+----------+----------+----------+
  |          |          |          |          | > lane p - 1
  +----------+----------+----------+----------+
Figure 9: Single-Pass Argon2 with p Lanes and 4 Slices
```

(Note: since slices are an important concept in Rust, to disambiguate I'll throw scare quotes on Argon2 "slices")

The parallel implementation is described as follows in the Argon2 paper:

https://www.password-hashing.net/argon2-specs.pdf

> We suggest the following solution for p cores: the entire memory is split into p lanes of l equal slices each,
which can be viewed as elements of a (p×l)-matrix Q[i][j]. Consider the class of schemes given by Equation (1).
>
> We modify it as follows:
> • p invocations to H run in parallel on the first column Q[∗][0] of the memory matrix. Their indexing
functions refer to their own slices only;
> • For each column j > 0, l invocations to H continue to run in parallel, but the indexing functions now may
[...]

PR #487, opened 2024-02-03
Reduce rust-version

pflanze:
Is there a reason why rust-version for argon2 was at 1.65? I'm compiling with rustc 1.63.0 from Debian bookworm, it compiled fine and I haven't seen any issues at runtime. The increase from 1.60 was done as part of #391 but I couldn't see a reasoning and it was done for the main Cargo.toml as well, maybe it was only relevant for the main crate.

### DIFF
--- a/.github/workflows/argon2.yml
+++ b/.github/workflows/argon2.yml
@@ -22,7 +22,7 @@ jobs:
     strategy:
       matrix:
         rust:
-          - 1.65.0 # MSRV
+          - 1.63.0 # MSRV
           - stable
         target:
           - thumbv7em-none-eabi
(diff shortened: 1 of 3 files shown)


pflanze:
Force-pushed to also set the MSRV for the CI.

pflanze:
And force-pushed again to also change it in the tests, as well as the docs.

tarcieri:
The main reason was to leave the door open to using GATs as part of a solution to #380, though granted no progress was made on that problem.

There's a high burden to maintaining older MSRVs in terms of dependency incompatibilities and overall project toil in that regard, longer compile times due to slower older compilers, and locking ourselves out of Rust features, and our upgrade cycles run quite long, so I'm just going to close this as I would prefer to upgrade, not downgrade, from MSRVs we commit to.

PR #490, opened 2024-03-20
Add PartialEq to scrypt::Params

uglyoldbob:
This allows for PartialEq structs that contain an scrypt Params to also be PartialEq.


newpavlov:
It also may be worth to derive `Eq` and `Hash`.

PR #500, opened 2024-05-17
sha-crypt: Change parameter type and make hashing infallible

### LINKED ISSUE #498: `Sha256Params::new`'s `round` parameter should probably not be a `usize`
[`Sha256Params::new`](https://docs.rs/sha-crypt/0.5.0/sha_crypt/struct.Sha256Params.html#method.new)

Since the number of rounds has nothing to do with the pointer width of the target CPU, a more concrete type like `u32` should probably be used. This is unfortunately a breaking change.

### LINKED ISSUE #499: `sha256_crypt` doesn't need an error return
[`sha256_crypt`](https://docs.rs/sha-crypt/0.5.0/sha_crypt/fn.sha256_crypt.html)

The only thing it uses the error return for is checking the passed `Sha256Params`, but those were already validated in `Sha256Params::new`.

tbu-:
Fixes #498.
Fixes #499.

### DIFF
--- a/sha-crypt/src/lib.rs
+++ b/sha-crypt/src/lib.rs
@@ -16,8 +16,7 @@
 //! let params = Sha512Params::new(10_000).expect("RandomError!");
 //!
 //! // Hash the password for storage
-//! let hashed_password = sha512_simple("Not so secure password", &params)
-//!     .expect("Should not fail");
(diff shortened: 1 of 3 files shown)


tarcieri (approved):

tarcieri:
Thanks!

PR #521, opened 2024-08-26
argon2: only enable password-hash dependency if its explicit

BlackHoleFox:
A missing weak feature specifier `?` was causing `password-hash` to always become part of a crate's build tree if someone had specified `default-features = false, features = ["alloc"]` in their `Cargo.toml`.

Due to https://github.com/rust-lang/cargo/issues/10801 `password-hash` will still always appear in the lockfile, but it won't be compiled anymore when a downstream crate has only enabled the `alloc` or `std` features.

### DIFF
--- a/argon2/Cargo.toml
+++ b/argon2/Cargo.toml
@@ -33,10 +33,10 @@ password-hash = { version = "0.6.0-rc.0", features = ["rand_core"] }
 
 [features]
 default = ["alloc", "password-hash", "rand"]
-alloc = ["password-hash/alloc"]
-std = ["alloc", "password-hash/std"]
+alloc = ["password-hash?/alloc"]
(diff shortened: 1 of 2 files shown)


tarcieri:
Looks like the documentation needs to be updated to be gated on `password-hash` and not just `std`: https://github.com/RustCrypto/password-hashes/blob/28dfc27/argon2/src/lib.rs#L33

(I guess when it isn't run under `cargo hack`, something else in the workspace is implicitly enabling the `password-hash` feature somehow)

tarcieri:
Needs rustfmt but looks like it's almost there

BlackHoleFox:
Oops, didn't realize the added features would line wrap.

tarcieri (approved):
Looks good now, thanks!

PR #533, opened 2024-10-07
scrypt: 0.12.0-pre.2

baloo:


baloo:
@tarcieri I saw you pushed the git tag, was there an issue in publishing the crate?

tarcieri:
Published!

PR #534, opened 2024-10-21
OSS-Fuzz: OSS-Fuzz fuzzing integration

arthurscchan:
Hi! Would you be interested in setting up fuzzing for the Scrypt module via OSS-Fuzz?

Fuzzing is essentially a stress-testing approach used to find bugs in software, and OSS-Fuzz is a free service run by Google that continuously fuzzes important open-source projects. Integrating your module with OSS-Fuzz could help uncover memory corruption issues that may exist.

This PR adds a Cargo fuzz configuration along with a fuzzer for the Scrypt module. In combination with an initial attempt in OSS-Fuzz (https://github.com/google/oss-fuzz/pull/12629), it enables OSS-Fuzz to fuzz the Scrypt module while keeping the fuzzers upstream for further modification and expansion. If you're happy to proceed with the integration and store the fuzzers upstream, please let me know, and I'd be glad to provide more details if needed.

The only thing required at this point is an email associated with a Google account, which will be used to receive notifications when bugs are found.


newpavlov:
Looks good! But I think it's better to have a `fuzz` crate in the repository root excluded from the main workspace.

arthurscchan:
Thanks for the suggestion. I have relocated the fuzz directory and fixes the Cargo.toml respectively.

Cargo.toml
```diff
@@ -4,6 +4,7 @@ members = [
     "argon2",
     "balloon-hash",
     "bcrypt-pbkdf",
+    "fuzz",
```
newpavlov, line 7: Exclude `fuzz` from the workspace, e.g. see the `benches` crate in the [stream cipher](https://github.com/RustCrypto/stream-ciphers) repository. It's not worth to pull the fuzz dependencies every time while testing the implementation crates. We can have a [separate CI job](https://github.com/RustCrypto/stream-ciphers/blob/master/.github/workflows/benches.yml) to check that `fuzz` compiles without errors.

tarcieri:
@arthurscchan we had some other projects added to oss-fuzz but the contributor disappeared. They now send me emails periodically to say it's broken and I don't have time to look into fixing it.

If we do this integration, are you interested in maintaining it going forward?

fuzz/Cargo.toml
```diff
@@ -0,0 +1,32 @@
...
+version = "0.0.0"
+publish = false
+edition = "2018"
+
+[package.metadata]
+cargo-fuzz = true
+
+[dependencies]
+libfuzzer-sys = "0.4"
+pbkdf2 = { version = "=0.13.0-pre.1", path = "../pbkdf2" }
```
newpavlov, line 12: I don't think you need all these dependencies. `libfuzzer-sys` and `scrypt` should be sufficient.

fuzz/Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+publish = false
+edition = "2018"
+
+[package.metadata]
+cargo-fuzz = true
+
+[dependencies]
+libfuzzer-sys = "0.4"
+scrypt = { path = "../scrypt" }
+password-hash = { version = "0.6.0-rc.0", default-features = false, features = ["rand_core"], optional = true }
```
newpavlov, line 13: You don't need to import `password-hash` explicitly. You can just use `scrypt = { path = "../scrypt", features = ["simple"] }` and then use [its re-export](https://github.com/RustCrypto/password-hashes/blob/master/scrypt/src/lib.rs#L65-L66) from `scrypt`.

fuzz/Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+hex = "0.4"
+base64 = "0.22.1"
+
+[features]
+default = ["simple", "std"]
+simple = ["password-hash"]
+std = ["password-hash/std"]
```
newpavlov, line 20: I think the `features` section can be removed, i.e. it's worth to enable the `simple` feature unconditionally.

fuzz/Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+
+[package.metadata]
+cargo-fuzz = true
+
+[dependencies]
+libfuzzer-sys = "0.4"
+scrypt = { path = "../scrypt" }
+password-hash = { version = "0.6.0-rc.0", default-features = false, features = ["rand_core"], optional = true }
+hex = "0.4"
+base64 = "0.22.1"
```
newpavlov, line 15: We have the [`base16ct`](https://docs.rs/base16ct) and [`base64ct`](https://docs.rs/base64ct) crates in our org, I would prefer to use them instead.

fuzz/Cargo.toml
arthurscchan: Sorry, I overlook the simple features in scrypt that re-export the password-hash crate. Fixed now.

fuzz/Cargo.toml
arthurscchan: Sorry, my bad, I forget to remove the unused base64 crate. Fixed now.

fuzz/Cargo.toml
arthurscchan: Agreed. Thanks for your suggestion. Fixed.

fuzz/Cargo.toml
newpavlov: Please replace the `hex` crate with `base16ct` as well.

arthurscchan: Fixed. Sorry for the confusion.

fuzz/fuzz_targets/scrypt_fuzzer.rs
```diff
@@ -0,0 +1,150 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#![no_main]
+
+use libfuzzer_sys::fuzz_target;
+use scrypt::{scrypt, Params};
+
+#[cfg(feature = "simple")]
```
newpavlov, line 20: These `cfg`s are no longer needed. Even worse, I think the crate will not even compile with these. Can you add a CI job for the `fuzz` crate, similar to the [`benches` job](https://github.com/RustCrypto/stream-ciphers/blob/master/.github/workflows/benches.yml)?

fuzz/fuzz_targets/scrypt_fuzzer.rs
```diff
@@ -0,0 +1,150 @@
+// Copyright 2024 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
```
newpavlov, line 3: Can you license this code under both Apache-2.0 and MIT? We also generally do not use copyright headers in source files and instead rely on the LICENSE files (you can put the Google copyright into LICENSE-MIT).

DavidKorczynski:
> @arthurscchan we had some other projects added to oss-fuzz but the contributor disappeared. They now send me emails periodically to say it's broken and I don't have time to look into fixing it.
> 
> If we do this integration, are you interested in maintaining it going forward?

We're happy to maintain it's building!

Which are the projects you're referring to, we could take a look at getting them build again.

tarcieri:
@DavidKorczynski the original PR was here: https://github.com/google/oss-fuzz/pull/6908

Here are the failures: https://issues.oss-fuzz.com/issues/42515273

fuzz/fuzz_targets/scrypt_fuzzer.rs
arthurscchan: I have fixed the license.

fuzz/fuzz_targets/scrypt_fuzzer.rs
arthurscchan: I have fixed the logic and import. I have also add a CI for testing the fuzz build with cargo fuzz.

.github/workflows/fuzz_build.yml
```diff
@@ -0,0 +1,32 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: RustCrypto/actions/cargo-cache@master
+      - uses: dtolnay/rust-toolchain@master
+        with:
+          toolchain: stable
+      - name: Install cargo-fuzz
+        run: |
+          cargo install cargo-fuzz
```
newpavlov, line 29: It's better to install a pre-compiled version, e.g. see how we [install `cross`](https://github.com/RustCrypto/actions/blob/master/cross-install/action.yml#L10-L13).

Also, maybe we don't need to install cargo-fuzz in the first place and just use `cargo build`?

fuzz/fuzz_targets/scrypt_fuzzer.rs
```diff
@@ -0,0 +1,141 @@
...
+        if !validate_salt(salt_string.as_str()) {
+            return;
+        }
+
+        let salt_value = salt_string.as_salt(); // Safe to use now
+
+        let formatted_hash = format!("$scrypt$ln=16,r=8,p=1${}$invalid$", base16ct::lower::encode_string(password));
+
+        if let Ok(hash) =
+            PasswordHash::new(SAMPLE_HASH).or_else(|_| PasswordHash::new(formatted_hash.as_str()))
```
newpavlov, line 104: I think it's better to write this part like this:
```rust
let hash = match PasswordHash::new(SAMPLE_HASH) {
   Ok(hash) => hash,
   Err(_) => return,
};

let salt_string = match fuzzed_salt(salt) {
    Some(s) if validate_salt(s) => s.as_salt(),
    _ => return,
};
```
IIUC `formatted_hash` is not needed at all, since parsing of `SAMPLE_HASH` should always succeed.

newpavlov:
@arthurscchan 
I've split the fuzz target into 3 separate targets. I think it results in a clearer code and slightly more efficient fuzzing. Please verify that I've translated your intention properly. Note that there are some minor differences, e.g. `SaltString` is created from raw bytes instead of parsing fuzer-provided B64 string.

fuzz/fuzz_targets/scrypt_phc_verify.rs
```diff
@@ -0,0 +1,13 @@
+#![no_main]
+use libfuzzer_sys::fuzz_target;
+use scrypt::password_hash::{PasswordHash, PasswordVerifier};
+use scrypt::Scrypt;
+
+const SAMPLE_HASH: &str = "$scrypt$ln=16,r=8,p=1$\
+    aM15713r3Xsvxbi31lqr1Q$nFNh2CVHVjNldFVKDHDlm4CbdRSCdEBsjjJxD+iCs5E";
+
+fuzz_target!(|password: &[u8]| {
+    let hash = PasswordHash::new(SAMPLE_HASH).expect("SAMPLE_HASH is valid");
```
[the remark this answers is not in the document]
arthurscchan: Thanks for the suggestion. I have tried before and I think the PasswordHash will throw panic with ABRT signal if invalid format is given. The fuzzing process will halt and create a false positive issue. For this reason I think given a sample PHC is better for fuzzing. Also, the main point is to fuzz the scrypt module, not the PasswordHash module in my understanding. Please correct me if I am wrong.

arthurscchan:
> @arthurscchan I've split the fuzz target into 3 separate targets. I think it results in a clearer code and slightly more efficient fuzzing. Please verify that I've translated your intention properly. Note that there are some minor differences, e.g. `SaltString` is created from raw bytes instead of parsing fuzer-provided B64 string.

Thanks for helping me to fix the fuzzers. I will double check the logic.

fuzz/fuzz_targets/scrypt_phc_verify.rs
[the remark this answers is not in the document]
newpavlov: I thought you wanted to fuzz our PHC implementation as well. I just don't think that `scrypt_phc_verify` checks that much in its current form, so it's probably better to just remove it.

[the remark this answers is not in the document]
arthurscchan: Please give me some moment, I will test again and see if the fuzzing for PHC implementation can be done. If yes, I could try changing it to random value and keep the fuzzing of PHC implementation as well.

newpavlov:
I merged the fuzz targets back into one. Now in addition to direct hashing using the `scrypt` function, it performs PHC hashing and verification of its result, i.e. instead of the fixed string we use serialized PHC hash for random password, salt, and parameters.

DavidKorczynski:
> @DavidKorczynski the original PR was here: [google/oss-fuzz#6908](https://github.com/google/oss-fuzz/pull/6908)
> 
> Here are the failures: https://issues.oss-fuzz.com/issues/42515273

Thanks!

We're happy to maintain this integration (password-hashes) but I think rustcrypto might be tricky because it relies on Cryptofuzz. @guidovranken is the right person to ask here -- @guidovranken do you about the build failures of rustcrypto?

PR #535, opened 2024-11-01
fuzz: add upper bound for scrypt's password and salt sizes

newpavlov:
Also change ranges used to generate scrypt parameters to prevent potential timeout issues.

cc @arthurscchan


newpavlov:
@arthurscchan
Can you close the two clusterfuzz issues or will it be done automatically?

arthurscchan:
@newpavlov Sorry for missing the size check. The OSS-Fuzz will update and verify the fix daily and close it automatically. Thanks for fixing the harness.

PR #544, opened 2024-12-10
feat(scrypt): Add `Params::n` method

sorairolake:
Currently, the `N` parameter can be obtained by raising 2 to the power of `Params::log_n`.

```rust
// example of raising 2 to the power of `Params::log_n`
let n = 1 << params.log_n();
```

I think it's easier to read to use a method which gets the `N` parameter directly instead of this way.

```rust
let n = params.n();
```

### DIFF
--- a/scrypt/src/params.rs
+++ b/scrypt/src/params.rs
@@ -102,9 +102,18 @@ impl Params {
 
     /// log₂ of the Scrypt parameter `N`, the work factor.
     ///
-    /// Memory and CPU usage scale linearly with `N`.
+    /// Memory and CPU usage scale linearly with `N`. If you need `N`, use
+    /// [`Params::n`] instead.
     pub const fn log_n(&self) -> u8 {
         self.log_n
+    }
+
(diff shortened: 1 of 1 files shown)


scrypt/src/params.rs
```diff
@@ -102,11 +102,20 @@ impl Params {
...
-    /// Memory and CPU usage scale linearly with `N`.
+    /// Memory and CPU usage scale linearly with `N`. If you need `N`, use
+    /// [`Params::n`] instead.
     pub const fn log_n(&self) -> u8 {
         self.log_n
     }
 
+    /// `N` parameter: the work factor.
+    ///
+    /// This method returns the square of [`Params::log_n`]. Memory and CPU
```
newpavlov, line 113: In my understanding, "the square of `log_n`" is `log_n^2`. I think it's better to write "2 to the power of `log_n`" or just `2^log_n`.

newpavlov (approved):

PR #567, opened 2025-03-04
Add BlockArray abstraction to handle fallible allocations

### LINKED ISSUE #566: `Argon2::hash_password_into` should use fallible memory allocations
`Argon2::hash_password_into` crashes on 32-bit architectures when using the FIRST RECOMMENDED parameter option (see [Section 4 of RFC 9106](https://www.rfc-editor.org/rfc/rfc9106.html#name-parameter-choice)) because it tries to infallibly allocate the 2 GiB buffer.

We believe that the straight-forward, easy-to-use interfaces should be safe by default.  Following that, `Argon2::hash_password_into` should make a fallible allocation, returning allocation errors instead of crashing.

conradludgate:
This resolves #566.

There was an open question regarding alignment. I couldn't demonstrate any significant performance benefit from unaligned allocations, so I skipped that here.

I wasn't sure what error to return. I decided not to introduce a breaking change here with a new error variant, but perhaps that is desired.

### DIFF
--- a/argon2/src/block.rs
+++ b/argon2/src/block.rs
@@ -154,3 +154,48 @@ impl Zeroize for Block {
         self.0.zeroize();
     }
 }
+
+/// `BlockArray` is like `Box<[Block]>`, but have our own abstraction that allows us to force alloc_zeroed.
+///
+/// This also allows easier access to fallible allocation.
+#[cfg(feature = "alloc")]
+pub(crate) struct BlockArray {
+    alloc: *mut u8,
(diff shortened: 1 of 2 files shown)


newpavlov:
Sorry, I had my own draft. I published it as #568. I borrowed some things from your PR (e.g. I did not know about `Layout::array`), but otherwise I think it's better to continue development of this feature there.

conradludgate:
No harm done :)

PR #547, opened 2025-01-13
argon2: add parallelism

jonasmalacofilho:
Adds a ~~default-enabled~~ `parallel` feature, with an ~~otherwise~~ optional dependency on `rayon`, and parallelize the filling of blocks using the memory views mentioned above.

Coordinated shared access in the memory blocks is implemented with a `SegmentViewIter` iterator, which implements either `rayon::iter::ParallelIterator` or `core::iter::Iterator` and returns `SegmentView` views into the Argon2 blocks memory that are safe to be used in parallel.

The views alias in the regions that are read-only, but are disjoint in the regions where mutation happens. Effectively, they implement, with a combination of mutable borrowing and runtime checking, the cooperative contract outlined in RFC 9106. This is similar to what was suggested in #380.

To avoid aliasing mutable references into the entire buffer of blocks (which would be UB), pointers are used up to the moment where a reference (shared or mutable) into a specific block is returned. At that point, aliasing is no longer possible.

The following tests have been tried in and pass Miri (modulo unrelated warnings):

    reference_argon2i_v0x13_2_8_2
    reference_argon2id_v0x13_2_8_2

(Running these in Miri is quite slow, taking ~5 minutes each, so I only ran the most obviously relevant tests for now).

~~Finally, the alignment of `Block`s increases to 128 bytes for better prevention of false sharing on modern platforms. The new value is based on notes on [`crossbeam-utils::CachePadded`](https://github.com/crossbeam-rs/crossbeam/blob/master/crossbeam-utils/src/cache_padded.rs#L63-L79).~~

---

I also took some inspiration from an [intermediate snapshot](https://github.com/RustCrypto/password-hashes/blob/73930b2cb619643f4bb306c7a0601d423e5eecc1/argon2/src/lib.rs) of #247, before the parallel implementation was removed, as well as from an implementation without any safe abstractions I just worked on for the rust-argon2 crate (sru-systems/rust-argon2#56).

PR #568, opened 2025-03-04
argon2: detect allocation failures in `hash_password_into`

### LINKED ISSUE #566: `Argon2::hash_password_into` should use fallible memory allocations

newpavlov:
Closes #566

### DIFF
--- a/argon2/src/block.rs
+++ b/argon2/src/block.rs
@@ -171,7 +171,7 @@ impl Blocks {
         if len == 0 {
             return None;
         }
-        
+
         let layout = Layout::array::<Block>(len).ok()?;
         // SAFETY: e use `alloc_zeroed` correctly
         let p = unsafe { alloc_zeroed(layout) };
--- a/argon2/src/block.rs
+++ b/argon2/src/block.rs
@@ -168,6 +168,10 @@ impl Blocks {
         use alloc::alloc::{Layout, alloc_zeroed};
         use core::ptr::NonNull;
 
+        if len == 0 {
+            return None;
+        }
+        
         let layout = Layout::array::<Block>(len).ok()?;
         // SAFETY: e use `alloc_zeroed` correctly
         let p = unsafe { alloc_zeroed(layout) };
--- a/argon2/src/block.rs
+++ b/argon2/src/block.rs
@@ -154,3 +154,43 @@ impl Zeroize for Block {
         self.0.zeroize();
     }
 }
+
+/// Custom implementation of `Box<[Block]>` until `Box::try_new_zeroed_slice` is stabilized.
+#[cfg(feature = "alloc")]
+pub(crate) struct Blocks {
+    p: core::ptr::NonNull<Block>,
+    len: usize,
+}
+
+#[cfg(feature = "alloc")]
+impl Blocks {
+    pub fn new(len: usize) -> Option<Self> {
+        use alloc::alloc::{Layout, alloc_zeroed};
+        use core::ptr::NonNull;
+
+        let layout = Layout::array::<Block>(len).ok()?;
+        // SAFETY: e use `alloc_zeroed` correctly
+        let p = unsafe { alloc_zeroed(layout) };
+
+        let p = NonNull::new(p.cast())?;
+        Some(Self { p, len })
+    }
+
+    pub fn as_slice(&mut self) -> &mut [Block] {
+        // SAFETY: `self.p` is a valid non-zero pointer that points to memory of the necessary size
+        unsafe { slice::from_raw_parts_mut(self.p.as_ptr(), self.len) }
+    }
+}
+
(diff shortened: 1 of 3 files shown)


argon2/src/error.rs
```diff
@@ -116,6 +120,7 @@ impl From<Error> for password_hash::Error {
             Error::ThreadsTooMany => InvalidValue::TooLong.param_error(),
             Error::TimeTooSmall => InvalidValue::TooShort.param_error(),
             Error::VersionInvalid => password_hash::Error::Version,
+            Error::AllocFail => InvalidValue::TooLong.param_error(),
```
newpavlov, line 123: I am not sure what to do here. Maybe we should add a new error variant to `password_hash::Error`?

argon2/src/block.rs
```diff
@@ -154,3 +154,43 @@ impl Zeroize for Block {
...
+
+#[cfg(feature = "alloc")]
+impl Blocks {
+    pub fn new(len: usize) -> Option<Self> {
+        use alloc::alloc::{Layout, alloc_zeroed};
+        use core::ptr::NonNull;
+
+        let layout = Layout::array::<Block>(len).ok()?;
+        // SAFETY: e use `alloc_zeroed` correctly
+        let p = unsafe { alloc_zeroed(layout) };
```
jonasmalacofilho, line 173: This depends on the caller (which, to be fair, is limited to this crate) not passing `len == 0`.

Wouldn't it be better to explicitly assert that `layout.size() > 0`?

argon2/src/error.rs
```diff
@@ -56,8 +56,11 @@ pub enum Error {
     /// Time cost is too small.
     TimeTooSmall,
 
-    /// Invalid version
+    /// Invalid version.
     VersionInvalid,
+
+    /// Failed to allocate memory.
+    AllocFail,
```
jonasmalacofilho, line 63: Note: this is a semver bump, as the enum is public and exhaustive.

jonasmalacofilho (commented):
Note that my allocation-only benchmark in #566 suggests this _could_ be a tiny bit slower than `vec![Block::default, ...]`.

So maybe it would be good to benchmark this with high-ish memory sizes, or at least with the 19 MiB and 46 MiB sizes recommended by OWASP?

argon2/src/block.rs
newpavlov: Ah, indeed. `alloc_zeroed` requires for `Layout` to be non-zero. We probably can simply return `None` in such case instead of using a dangling pointer on zero length.

tarcieri:
Maybe we should try to land #547 first so we can gauge any impact on performance?

argon2/src/error.rs
tarcieri: Sure

argon2/src/block.rs
```diff
@@ -154,3 +154,47 @@ impl Zeroize for Block {
...
+    pub fn new(len: usize) -> Option<Self> {
+        use alloc::alloc::{Layout, alloc_zeroed};
+        use core::ptr::NonNull;
+
+        if len == 0 {
+            return None;
+        }
+        
+        let layout = Layout::array::<Block>(len).ok()?;
+        // SAFETY: e use `alloc_zeroed` correctly
```
tarcieri, line 176: I'm not sure what this is trying to say?

argon2/src/error.rs
```diff
@@ -79,6 +82,7 @@ impl fmt::Display for Error {
             Error::ThreadsTooMany => "too many threads",
             Error::TimeTooSmall => "time cost is too small",
             Error::VersionInvalid => "invalid version",
+            Error::AllocFail => "allocation failure",
```
tarcieri, line 85: Perhaps `OutOfMemory`?

argon2/src/block.rs
newpavlov: Fixed.

newpavlov:
>Maybe we should try to land https://github.com/RustCrypto/password-hashes/pull/547 first so we can gauge any impact on performance?

I think we should merge this PR, then #547 without the alignment workaround, and then finally add the alignment workaround to `Blocks` in a separate PR. I am still not 100% convinced that we should add the workaround. It needs additional investigation in my opinion.

tarcieri (approved):

PR #572, opened 2025-03-05
Update Cargo.lock

newpavlov:
Using the yanked version of `base64ct` results in a security audit warning:
https://github.com/RustCrypto/password-hashes/actions/runs/13675565491/job/38235226692

### DIFF
--- a/argon2/src/error.rs
+++ b/argon2/src/error.rs
@@ -123,6 +123,8 @@ impl From<Error> for password_hash::Error {
 impl core::error::Error for Error {
     fn source(&self) -> Option<&(dyn core::error::Error + 'static)> {
         match self {
+            // TODO: remove afer `B64Error` will implement `core::error::Error`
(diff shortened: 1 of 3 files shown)


argon2/src/error.rs
```diff
@@ -123,6 +123,8 @@ impl From<Error> for password_hash::Error {
 impl core::error::Error for Error {
     fn source(&self) -> Option<&(dyn core::error::Error + 'static)> {
         match self {
+            // TODO: remove afer `B64Error` will implement `core::error::Error`
+            #[cfg(feature = "std")]
```
newpavlov, line 127: This is a consequence of yanking `base64ct` v1.7 (see https://github.com/RustCrypto/formats/issues/1684).

@tarcieri 
What are you thoughts on `base64ct`'s MSRV? Should we release v2 with the bump?

PR #591, opened 2025-05-11
chore(deps): bump `crypto-bigint` to `0.7.0-pre.2`

baloo:


tarcieri:
Just released a `pre.3` as it were

PR #606, opened 2025-05-30
scrypt: fixup doc compilation

baloo:
When scrypt used a dependency with `default-features = false`, the doc compilation will fail because `password_hash::PasswordHasher` will be unavailable (`password-hash` is only pulled `simple` feature is used).

It did not show up in CI, because `password-hash` is also provided as a `dev-dependency` and is available when `cargo test` evaluates the documentation.

Fixup #602

### DIFF
--- a/.github/workflows/scrypt.yml
+++ b/.github/workflows/scrypt.yml
@@ -60,3 +60,4 @@ jobs:
       - run: cargo test --no-default-features
       - run: cargo test
       - run: cargo test --all-features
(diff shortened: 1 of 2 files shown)


baloo:
@tarcieri any chance to make another release of scrypt after this?

https://github.com/RustCrypto/formats/pull/1847 for the downstream build that failed

tarcieri (approved):

PR #633, opened 2025-09-04
yescrypt trivial changes

conradludgate:
Fairly trivial changes for the c2rust yescrypt code to become a bit more rusty. I have some more substantial changes on the way.


yescrypt/src/lib.rs
```diff
@@ -164,23 +142,22 @@ pub fn yescrypt_kdf(
...
-    let mut flags: Flags = (*params).flags;
-    let mut N: uint64_t = (*params).N;
-    let mut r: uint32_t = (*params).r;
-    let mut p: uint32_t = (*params).p;
-    let mut t: uint32_t = (*params).t;
-    let mut g: uint32_t = (*params).g;
-    let mut NROM: uint64_t = (*params).NROM;
+    let flags: Flags = (*params).flags;
+    let N: uint64_t = (*params).N;
+    let r: uint32_t = (*params).r;
```
tarcieri, line 156: For future reference all the superfluous syntax here is easy to get rid of:

```suggestion
    let flags = params.flags;
    let N = params.N;
    let r = params.r;
```

Not a blocker for this PR though, you're already doing a lot already!

yescrypt/src/lib.rs
```diff
@@ -355,74 +313,15 @@ unsafe fn yescrypt_kdf_body(
                                 match current_block {
                                     15162489974460950378 => {}
                                     _ => {
-                                        VROM = ptr::null();
-                                        if !shared.is_null() {
```
tarcieri, line 359: I guess you got rid of `shared` and wow, that simplifies things

PR #634, opened 2025-09-04
yescrypt: significant changes to control flow for kdf_body

conradludgate:
Stacked on top of #633 

This takes several steps to try and disentagnle all the states inside of the kdf_body. As we move to RAII, we should be able to get rid of most of this structure and just rely on drop guards to kick in.


yescrypt/src/lib.rs
```diff
@@ -282,539 +234,301 @@ unsafe fn yescrypt_kdf_body(
...
+                let mut dkp = buf;
+                if flags != 0 && buflen < size_of::<[uint8_t; 32]>() {
+                    PBKDF2_SHA256(
+                        passwd,
+                        passwdlen,
+                        B as *mut uint8_t,
+                        B_size,
+                        1 as libc::c_int as uint64_t,
+                        dk.as_mut_ptr(),
+                        size_of::<[uint8_t; 32]>(),
```
tarcieri, line 442:
```suggestion
                        32,
```

PR #643, opened 2025-09-04
yescrypt: use `bitflags` crate for flags

tarcieri:
It was previously hard to tell what was happening with flags because they were all hardcoded constants after the C-to-Rust translation.

This uses the `bitflags` crate as a first pass towards trying to clean them up and make them more semantic so we can tell what's actually happening.

cc @conradludgate

### DIFF
--- a/yescrypt/src/lib.rs
+++ b/yescrypt/src/lib.rs
@@ -68,9 +68,23 @@ struct Local {
     pub aligned: Box<[u32]>,
 }
 
-type Flags = u32;
+bitflags::bitflags! {
+    #[derive(Clone, Copy, Debug, Eq, PartialEq)]
+    pub struct Flags: u32 {
+        const WORM        = 0x001;
+        const RW          = 0x002;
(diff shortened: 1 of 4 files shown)


tarcieri:
Note: this PR doesn't add or convert all of the flags yet, but it's a start

yescrypt/src/lib.rs
```diff
@@ -68,9 +68,23 @@ struct Local {
...
+    pub struct Flags: u32 {
+        const WORM        = 0x001;
+        const RW          = 0x002;
+        const ROUNDS_3    = 0b000;
+        const ROUNDS_6    = 0x004;
+        const GATHER_4    = 0x010;
+        const SIMPLE_2    = 0x020;
+        const SBOX_12K    = 0x080;
+        const INIT_SHARED = 0x01000000;
+        const ALLOC_ONLY  = 0x08000000;
```
tarcieri, line 82: We can probably get rid of these?

conradludgate (approved):
Nice!

PR #700, opened 2025-09-09
`pbkdf2`: remove unnecessary bounds

daxpedda:
I believe these bounds were only necessary in the older `hmac` versions.

### DIFF
--- a/pbkdf2/src/lib.rs
+++ b/pbkdf2/src/lib.rs
@@ -102,14 +102,7 @@ use rayon::prelude::*;
 use digest::{FixedOutput, InvalidLength, KeyInit, Update, typenum::Unsigned};
 
 #[cfg(feature = "hmac")]
-use {
-    digest::{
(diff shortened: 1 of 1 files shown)


pbkdf2/src/lib.rs
```diff
@@ -230,9 +223,7 @@ where
 #[cfg(feature = "hmac")]
 pub fn pbkdf2_hmac<D>(password: &[u8], salt: &[u8], rounds: u32, res: &mut [u8])
 where
-    D: EagerHash + HashMarker + Update + FixedOutput + Default + Clone,
-    <D as EagerHash>::Core: Sync,
-    <D as BlockSizeUser>::BlockSize: IsLess<U256, Output = True> + NonZero,
+    D: EagerHash<Core: Sync>,
```
newpavlov, line 226: Huh, I did not know we can specify bounds in such way.

PR #716, opened 2025-10-15
yescrypt v0.1.0-rc.0

tarcieri:


tarcieri:
Changed my mind: we should try to make `password-hash` generic with MCF support before releasing `.rc*`

PR #737, opened 2025-11-05
Bump `password-hash` to v0.6.0-rc.2

tarcieri:
This upgrades `rand_core` to v0.10 prereleases and depends directly on `getrandom` for access to the system RNG.

It additionally bumps other crates like `digest` and `blake2` which use the new `crypto-common` with an upgraded `rand_core`.

This required restoring some functionality in `blake2` which was used by `argon2`: https://github.com/RustCrypto/hashes/pull/754

This is currently referenced as a git branch until we decide if this is a permanent solution for `argon2` or not, so as to unblock the upgrade.


tarcieri:
Mostly I want to get new versions of `bcrypt-pbkdf2`, `pbkdf2`, and `scrypt` out here

PR #827, opened 2026-01-21
ci: update Clippy to 1.92

newpavlov:
Also change Clippy flags to `--workspace --all-features --lib --bins --tests` for better coverage.

### DIFF
--- a/scrypt/src/lib.rs
+++ b/scrypt/src/lib.rs
@@ -17,8 +17,14 @@
 //!
 //! # Usage (simple PHC password hash usage with default params)
 //!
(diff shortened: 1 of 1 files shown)


tarcieri:
If it has criterion benchmarks, I don't think it makes sense to keep the nightly-only benchmarks that are much less accurate.

Sidebar: I continue to find the `benches` crate stuff confusing.

newpavlov:
`scrypt` is fixed in #828.

>The cost of compiling criterion is minimal

On my notebook compiling just `criterion` (without benches) takes ~7 seconds and this time will be added to every job. It also will be felt during local builds for every dependency tree re-compilation. We can discuss whether it's worth or not, but it's better to do it in a separate issue.

PR #853, opened 2026-03-19
pbkdf2: remove `sha1` feature

tarcieri:
This came up in some discussions about hypothetical FIPS support, noting NIST already recommends against SHA-1 and plans on complete phase-out by December 31st, 2030:

https://www.nist.gov/news-events/news/2022/12/nist-retires-sha-1-cryptographic-algorithm

That article notes on that date FIPS 180-5 will be published which removes SHA-1, and a revision to SP 800-131A will reflect that SHA-1 is unsuitable for use in FIPS modules:

> “Modules that still use SHA-1 after 2030 will not be permitted for purchase by the federal government,” Celi said. “Companies have eight years to submit updated modules that no longer use SHA-1. Because there is often a backlog of submissions before a deadline, we recommend that developers submit their updated modules well in advance, so that CMVP has time to respond.” 

Though we're not quite to that deadline, this PR proposes to remove direct support for SHA-1 in anticipitation of these changes.

### DIFF
--- a/pbkdf2/tests/pbkdf2.rs
+++ b/pbkdf2/tests/pbkdf2.rs
@@ -2,6 +2,7 @@
 
 use belt_hash::BeltHash;
 use hex_literal::hex;
+use sha1::Sha1;
 use streebog::Streebog512;
 
 macro_rules! test {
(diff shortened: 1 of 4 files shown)


pbkdf2/tests/pbkdf2.rs
```diff
@@ -19,23 +19,6 @@ macro_rules! test {
     };
 }
 
-/// Test vectors from RFC 6070:
-/// https://www.rfc-editor.org/rfc/rfc6070
-#[test]
-fn pbkdf2_rfc6070() {
```
newpavlov: I think we can keep this test.

tarcieri:
The algorithm-specific support is needed/useful for implementing MCF and PHC support, which is only defined for PBKDF2 when used with HMAC-SHA1 and HMAC-SHA2

newpavlov:
Yes, but it does not mean that it should be part of `pbkdf2`. One potential solution could be to move it into a separate "composite" crate (e.g. `pbkdf2-sha2` or such), it also would help with the compatible versions selection issue.

newpavlov:
PBKDF2 is not a NIST-only algorithm (unlike you argued with GCM) and it's defined generically. Sure, SHA-2 it the most popular choice, but the same could be said about AES with the CTR mode. And I am _surely_ do not think we should include `aes` as an optional dependency of `ctr`.

newpavlov:
What makes you say that someone would not like to use PBKDF2 with Streebog or  BLAKE2/3 in MCF/PHC contexts? IIUC there is no OID-like centralized registry for algorithm IDs used in MCF/PHC, names like `pbkdf2-sha256` are mostly historic convention. Theoretically an app could use a completely different name if it wants. The [spec](https://github.com/P-H-C/phc-string-format/blob/master/phc-sf-spec.md) does not say anything about how `<id>` should be assigned.

PR #865, opened 2026-03-24
sha-crpypt, ballon-hash, bcrypt-pbkdf, and yescrypt benchmarks

WingZer0o:
Sha-crypt was missing some benchmarks. I propose that all of the individual password hashing crates have a few benchmarks in them based off the test vectors that are being put in. I can do the rest in a subsequent PR.


tarcieri:
@WingZer0o sorry, it's not that these aren't useful but they're kind of a low priority right now

WingZer0o:
@tarcieri Ah no worries, I wasn't sure so I just closed it. I have reopened and will fix the merge conflicts at some point.

PR #888, opened 2026-04-21
Update copyright year to 2026

tarcieri:


bcrypt-pbkdf/LICENSE-MIT
```diff
@@ -1,5 +1,5 @@
 Copyright (c) 2019 Jack Grigg
-Copyright (c) 2019-2023 The RustCrypto Project Developers
+Copyright (c) 2019-2024 The RustCrypto Project Developers
```
newpavlov: BTW what order is usually used? Here we order entries from older to newer, but in some other crates we do it in reverse.

Also, use 2026 instead of 2024?

tarcieri: I'm used to them being listed in-order

PR #884, opened 2026-04-20
pbkdf2: apply workspace-level lints

tarcieri:
Adds the workspace-level config from RustCrypto/utils#1411 to this repo and applies it to `pbkdf2`.

PR #892, opened 2026-04-21
scrypt: enable and fix workspace-level lints

tarcieri:
Enables and fixes the workspace-level lints added in #884, which is a standard configuration we're using across repositories

### DIFF
--- a/scrypt/src/block_mix.rs
+++ b/scrypt/src/block_mix.rs
@@ -3,22 +3,29 @@ cfg_if::cfg_if! {
         mod pivot;
         mod simd128;
         pub(crate) use simd128::{scrypt_block_mix, shuffle_in, shuffle_out};
+
+        #[cfg(test)]
+        #[path = "block_mix/soft.rs"]
+        mod soft_test;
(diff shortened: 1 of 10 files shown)


scrypt/src/block_mix.rs
```diff
@@ -3,40 +3,54 @@ cfg_if::cfg_if! {
...
         pub(crate) fn shuffle_out(_input: &mut [u8]) {}
+
+        #[cfg(test)]
+        use soft as soft_test;
     }
 }
 
-#[cfg(test)]
-#[path = "block_mix/soft.rs"]
-mod soft_test;
```
tarcieri, line 21: This approach was importing it twice

scrypt/src/params.rs
```diff
@@ -96,8 +102,8 @@ impl Params {
 
         Ok(Params {
             log_n,
-            r: r as u32,
-            p: p as u32,
```
tarcieri, line 100: Previously it was being cast from `u32` to `usize` and back to `u32`. While there are checks that should work, it seems more straightforward not to cast it back and forth when we're given and want a `u32`.

PR #905, opened 2026-06-01
widen s_cost * p_cost to usize when sizing parallel memory

dxbjavid:
With the parallel feature, hash_password_into and balloon_m size the per-thread memory as s_cost * p_cost evaluated in u32 and only then cast to usize. Both come from NonZeroU32 params and can be parsed from a hash string, so a product reaching 2^32 wraps: release builds get an undersized (often zero-length) buffer and a passing MemoryTooLittle check, debug builds panic on the multiply. Casting each operand to usize first computes the real size on 64-bit targets.

### DIFF
--- a/balloon-hash/src/balloon.rs
+++ b/balloon-hash/src/balloon.rs
@@ -49,7 +49,7 @@ where
     let output_xor = {
         use rayon::iter::{ParallelBridge, ParallelIterator};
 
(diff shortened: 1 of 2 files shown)


dxbjavid:
Hi team,

Just checking in on this PR. I wanted to see if there are any updates regarding the review or next steps.

Please let me know if any further changes are required from my side.

Thank you!

dxbjavid:
Closing this pull request due to inactivity

PR #908, opened 2026-06-02
widen r * p to u64 in scrypt params bounds check

dxbjavid:
Noticed the final guard in `Params::new` computes `r * p` in u32, but both come from a parsed PHC hash string and the earlier checked_mul guards only bound the usize products. On 64-bit, r=65536 p=65536 reaches this line: the u32 multiply wraps to 0, so `r * p >= 0x4000_0000` is false and scrypt accepts params it must reject (debug builds panic on the overflow instead). Widen both operands to u64 before the compare.

### DIFF
--- a/scrypt/src/params.rs
+++ b/scrypt/src/params.rs
@@ -96,7 +96,7 @@ impl Params {
         // check: p <= ((2^32-1) * 32) / (128 * r)
         // It takes a bit of re-arranging to get the check above into this form,
         // but it is indeed the same.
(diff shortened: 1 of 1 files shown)


dxbjavid:
Hi team,

Just checking in on this PR. I wanted to see if there are any updates regarding review or next steps.

Please let me know if any further changes are required from my side.

Thank you!

dxbjavid:
Closing this pull request due to inactivity

tarcieri:
Would still like this

PR #933, opened 2026-09-05
argon2: wipe intermediate buffers

karalabe:
Wipes various intermediate buffers. Also does a `mut` addition, otherwise we're passing and wiping a copy of `initial_hash`, but not the original.


newpavlov:
Relevant [comment](https://github.com/RustCrypto/KEMs/pull/366#issuecomment-5034435499).

@tarcieri 
WDYT our policy should be in this regard? It may be worth to open a meta issue for this.

karalabe:
Hmm, yes, it's a can of worms

tarcieri:
Yeah, perhaps we should open a meta-issue, but my inclination would continue to be to wipe the stack in a single pass at the end of the operation, instead of zeroing buffers as-you-go. It should be a lot more foolproof, because as your comment notes Rust makes it very easy to accidentally make copies of things.

newpavlov:
Opened https://github.com/RustCrypto/meta/issues/38 to discuss this.

newpavlov:
Closing per discussion in https://github.com/RustCrypto/meta/issues/38.

